PR #35, opened 2023-02-01
WIP: refact, use  use dav1d style HIGHBD_DECL_SUFFIX to reduce code size

nuomi2021:


nuomi2021:
This may not be worth it. Because 8 bits and 10 bits will use different functions.  16 bits may need 32 bits internal data format. 
so it cannot share functions with other bits. only 10 and 12 bits may share the same function. 
Let's close and revisit this later.

PR #48, opened 2023-03-11
workflow: add checkasm test

nuomi2021:


nuomi2021:
AVX2 is not supported. see https://github.com/actions/runner/issues/1069
close it

Issue #50, opened 2023-03-11
Refact Executor to reduce shedule cost


nuomi2021:
The current executor will use a single queue for not-ready and ready tasks. 
So, every task posted to the queue may wake all threads. 
Better split it into two queues: 1 for not ready tasks. 1 for ready tasks. it is protected by different locks
The thread adds a task to the executor who will also check readiness for all tasks in the not-ready queue.
once it is ready it will add to the ready queue.

the target is reduce the lock an unlock usage in perf top result
11.96% ffmpeg_g [.] put_vvc_luma_hv_10
5.88% ffmpeg_g [.] alf_get_coeff_and_clip_10
5.25% ffmpeg_g [.] ff_vvc_inv_dct2_64
4.30% [kernel] [k] __lock_text_start
4.22% ffmpeg_g [.] ff_vvc_alf_filter_luma_w16_16bpc_avx2
3.46% ffmpeg_g [.] put_vvc_luma_bi_hv_10
3.45% ffmpeg_g [.] alf_filter_luma_vb_10
3.13% ffmpeg_g [.] vvc_loop_filter_luma_10
2.81% ffmpeg_g [.] lmcs_filter_luma_10
2.46% ffmpeg_g [.] put_vvc_luma_uni_hv_10
2.27% ffmpeg_g [.] put_vvc_chroma_hv_10
2.21% libc-2.31.so [.] 0x000000000018b733
2.05% libc-2.31.so [.] 0x000000000018bb41
1.95% ffmpeg_g [.] put_vvc_chroma_uni_hv_10
1.84% ffmpeg_g [.] put_vvc_chroma_bi_hv_10
1.81% ffmpeg_g [.] vvc_deblock_bs
1.41% ffmpeg_g [.] ff_vvc_predict_inter
**1.25% libpthread-2.31.so [.] __pthread_mutex_lock
1.24% libpthread-2.31.so [.] __pthread_mutex_unlock**
1.22% ffmpeg_g [.] ff_vvc_residual_coding
1.08% ffmpeg_g [.] alf_filter_cc_10
1.03% ffmpeg_g [.] apply_prof_uni_10
0.99% ffmpeg_g [.] ff_vvc_alf_filter
0.98% ffmpeg_g [.] ff_vvc_inv_dct2_32
0.94% ffmpeg_g [.] vvc_deblock_bs_luma_vertical
0.92% ffmpeg_g [.] add_residual_10

[...]

PR #54, opened 2023-04-01
executor.c: Refactor to use two queues.

danielsan901998:
Implement #50
The queue is divided in two, one for ready and the other for not ready tasks, it use a new lock for the new queue.
In ff_executor_execute the ready status is checked to decide in what queue to insert, and only use pthread_cond_signal for the ready queue.
To update the not_ready queue a new function called update_tasks is executed after each runned task.
In update_tasks the lock of notready_tasks is acquired while iterating the queue, checking the ready status and removing the tasks from the queue and adding to a temporal queue called task_buff used as buffer to save the tasks to add to the ready queue, after the iteration task_buff is checked to see if there are task to add, if there are tasks the ready_lock is acquired and for each task we use priority to find when to insert it.
Result of executing the conformance tests:
`total = 211, passed = 211, failed = 0, skipped = 0`
`total = 62, passed = 0, failed = 3, skipped = 59`
I tried to using `perf record python3 tests/tools/ffmpeg.py --threads 1 tests/performance` to see a difference in __pthread_mutex_lock but in my machine i see:
`0,08%  ffmpeg_g  [kernel.kallsyms]  [k] native_queued_spin_lock_slowpath.part.0`
I'm not sure if i'm executing an incorrect command or if my CPU or Linux Kernel is causing the difference, so i don't have information on how efficient is my code to reduce schedule cost.

### DIFF
--- a/libavcodec/executor.c
+++ b/libavcodec/executor.c
@@ -35,10 +35,12 @@ struct Executor {
     uint8_t *local_contexts;
     int thread_count;
 
-    pthread_mutex_t lock;
+    pthread_mutex_t notready_lock;
+    pthread_mutex_t ready_lock;
     pthread_cond_t cond;
     int die;
-    Task *tasks;
+    Task *notready_tasks;
+    Task *ready_tasks;
 };
 
 static void remove_task(Task **prev, Task *t)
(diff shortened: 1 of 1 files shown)


nuomi2021:
@danielsan901998 , thank you for the patch.
It's too complex, how about we do things like this
1.  in ff_executor_execute, 
a. hold the lock for pending list
b. add the task to a pending list 
c. look at the tasks in the pending list, if it's ready, add it to a temporary list.
d. unlock the pending list.
e. lock the ready list
f. append the temporary list to the ready list, using priority.
g unlock the ready list and notify worker threads
2. in the work thread function, we do not need to change anything

Please also provide fps before and after your patch. thank you

danielsan901998:
I have tried it, but updating the pending list in ff_executor_execute does not works in my implementation.
Using gdb i see that the threads running executor_worker_task are waiting in ff_vvc_await_progress.
Updating the pending list after running a task is the only option i found to guarantee that the pending list is always updated and progress is always happening.

FPS results using the command `ffmpeg_g -i tests/performance/[video]  -vsync 0 -f rawvideo /dev/null -y` using the patch submitted:
| video | before | after |
| --------------- | --------------- | --------------- |
| BQTerrace_1920x1080_60_10_420_22_RA.vvc | 42 | 42
| Chimera_8bit_1080P_1000_frames.vvc | 75 | 76
| RitualDance_1920x1080_60_10_420_32_LD.266 | 65 | 64
| RitualDance_1920x1080_60_10_420_37_RA.266 | 72 | 73
| Tango2_3840x2160_60_10_420_27_LD.266 | 13 | 13

My CPU is AMD Ryzen 7 5800H

nuomi2021:
@danielsan901998  you are right, the parser and inter preidct's is_ready function is fake. if we add all parser and inter predict tasks to ready list, it will use up executors. I will remove the first good issue label for https://github.com/ffvvc/FFmpeg/issues/50
Thank you for your findings

Issue #55, opened 2023-04-02
Windows multithreading test failure


frankplow:
Had some time so did a little bit more research on the Windows CI (#52) [test failure](https://github.com/ffvvc/FFmpeg/actions/runs/4502421677/jobs/7924210755).

memsetting the entire `lc->sao_buffer` like bab47cad2c2b6d78a17765161a853dd2dcd46775 does not fix the issue, so I don't think the issue is related to #26. With this change, valgrind and clang's address sanitiser don't report any memory issues.

I have compiled FFmpeg directly with MSVC/MSYS2 (i.e. not via FFVS-Project-Generator) and the problem is similar so I don't think it's anything to do with the build files. I haven't yet got the gcc/MSYS2 toolchain or MinGW gcc cross-compilation working unfortunately.

I can't get the LTRP_A_ERICSSON_3 failure to reproduce on my machine, so I don't think there's anything special about this test. The tests which fail most frequently on my machine are:
* LMCS_A_Dolby_3.bit
* WPP_B_Sharp_2.bit

The failures only occur when running tests concurrently, they do not occur when running the tests individually or when running tests using a single thread. I don't know whether this points towards [libavcodec/vvc_thread.c](https://github.com/ffvvc/FFmpeg/blob/main/libavcodec/vvc_thread.c) at all? ~~[This line](https://github.com/ffvvc/FFmpeg/blob/b617f5c0f9c4944b83b57d674eae4af0e3298a4f/libavcodec/vvc_thread.c#L572) is part of what is preventing cross-compilation at the moment. Should it not be testing for compiler using `_MSC_VER` or something instead of checking the OS?~~ See #57 for fix. Don't believe this is related.

frankplow:
This issue is partially due to the lack of atomic operations for 8-bit types with MSVC winnt.h. Fixing this this will require an upstream change (see patch [here](https://patchwork.ffmpeg.org/project/ffmpeg/patch/20230405152629.35740-1-post@frankplowman.com/)) and then changing `avails` to a `atomic_uint *`.

With these patches + bab47ca applied, the errors are mostly gone except for LTRP_A_ERICSSON_3 – maybe there is something special about this test case after all? I can now reproduce the errors when a single test is run, rather than as a part of the suite and the decoded MD5 is different each time. MSYS2 does not have valgrind unfortunately. I might try generating a VS solution with FFVVS-Project-Generator and debugging with VS – I see how that's handy already!

nuomi2021:
[...]

PR #56, opened 2023-04-02
Change Win32

Jamaika1:
```
In file included from vvcdec.h:27,
                 from vvc_thread.h:26,
                 from vvc_thread.c:23:
vvc_thread.c: In function 'ff_vvc_task_run':
vvc_thread.c:577:13: error: size mismatch in argument 2 of '__atomic_compare_exchange'
  577 |             atomic_compare_exchange_strong(&ft->ret, &zero, ret);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/libavcodec/vvc_thread.c
+++ b/libavcodec/vvc_thread.c
@@ -569,7 +569,7 @@ int ff_vvc_task_run(Task *_t, void *local_context, void *user_data)
 
     if (!atomic_load(&ft->ret)) {
         if ((ret = run[t->type](s, lc, t)) < 0) {
-#ifdef WIN32
+#ifndef WIN32
             intptr_t zero = 0;
 #else
             int zero = 0;


frankplow (changes requested):
This should probably be
```c
#ifdef _MSC_VER
```

frankplow:
Think this is meant to be a fix for the issues with MinGW cross-compilation mentioned in #55. This allows MinGW to compile (not run without errors though) but breaks compilation on gcc for Linux and introduces a warning with MSVC in the process.

libavcodec/vvc_thread.c
frankplow, line 1: This should probably be
```
#ifdef _MSC_VER
```

Jamaika1: > Think this is meant to be a fix for the issues with MinGW cross-compilation mentioned in #55. This allows MinGW to compile (not run without errors though) but breaks compilation on gcc for Linux and introduces a warning with MSVC in the process.

Hi fankplow

I'm an amateur who got annoyed on the doom9 forum with a lot of vvc ads and no specific types of examples of converters made.
So I decided to manually make ffmpeg with vvc, evc, htj2k, jxs, jxl, avs, avs2, avs3 converter and see if it works and how it works.
There were problems because many functions overlap from different codecs and some small bugs that I tried to fix.
It took me 1.5 months.
I converted ultravideo kvazaar addon into vvc uvg266. FFmpeg makes codecs, but only vvc, mts.
The problem is they don't play. Non-standard codec?
As for Elecard's vvc bin advertising files. Only some frames are displayed.
gcc issues and the __atomic function.
From what I can see the GNOME based librsvg addon has dropped gcc compatibility with WIN32. As a non-startable function, only MSVC is usable. So the rsvg decoder was trashed.
https://gitlab.gnome.org/GNOME/glib/-/issues/600

nuomi2021:
> Hi @Jamaika1
> 
> > FFmpeg makes codecs, but only vvc, mts. The problem is they don't play. Non-standard codec?
> 
> VVC is a recent codec and so it is not supported in many players yet, including ffplay and VLC. It very much is standardised though.
Please work closely with us, once our code lands on FFmpeg, the entire world will get the vvc decoder :)

nuomi2021:
@Jamaika1  replaced by https://github.com/ffvvc/FFmpeg/pull/97

nuomi2021:
@Jamaika1 , could you help confirm the fix and close this?
thank you

Issue #26, opened 2023-01-25
valgrind issue for SAO filter.


nuomi2021:
run "valgrind ./ffmpeg_g -i SLICES_A_HUAWEI_3.bit -f rawvideo /dev/null -y"

you will get

> ==5154== Use of uninitialised value of size 8
> ==5154==    at 0x46E433: sao_edge_filter_10 (vvcdsp_template.c:200)
> ==5154==    by 0x50ACE9: ff_vvc_sao_filter (vvc_filter.c:309)
> ==5154==    by 0x52C1DB: run_sao (vvc_thread.c:447)
> ==5154==    by 0x52B462: ff_vvc_task_run (vvc_thread.c:567)
> ==5154==    by 0x500AE9: executor_worker_task (executor.c:79)
> ==5154==    by 0x49C8608: start_thread (pthread_create.c:477)
> ==5154==    by 0x4B02132: clone (clone.S:95)

It's realted to sao temporary buffer lc->sao_buffer, once you memset the buffer [here ](https://github.com/ffvvc/FFmpeg/blob/b4e85c10055500f8553d36f24b69399469387234/libavcodec/vvc_filter.c#L225), valgrind will not report any issue.

light2802:
Hi @nuomi2021. My name is Aarya Chaumal and I would like to take this issue up as my qualification task for this year's GSoC any advice on how should I set up my development environment for this task? I'm on a Linux machine.

nuomi2021:
great! you can check https://github.com/ffvvc/FFmpeg/blob/main/.github/workflows/makefile.yml for configuration and bitstream download. 
thank you

nuomi2021:
@light2802 is everything smooth?
thank you

light2802:
I sent the patch to a ffmpeg-devel mailing list but it is not showing up on the patchwork.ffmpeg.org/ page

nuomi2021:
the code is not merged by ffmpeg yet. could you help send pr to https://github.com/ffvvc/FFmpeg instead?

thank you

light2802:
Sure created a PR at https://github.com/ffvvc/FFmpeg/pull/49

nuomi2021:
remove low priority since #64

PR #64, opened 2023-04-15
Add Windows builds with assembly optimisations enabled to CI

frankplow:
This PR adds two new jobs to CI for building and testing with assembly optimisations enabled and assembled with NASM and YASM on Windows.

Different operating systems are now combined into a single matrix as this is easier to manage with all the combinations of OS/compiler/assembler. I think this will also be more future-proof as different checks and combinations are added. This required using MSYS/MSVC instead of FFVS-Project-Generator. If we still want to check FFVS-Project-Generator can generate a valid VS solution in CI I can add this as a separate job.

The new jobs have flagged an [error when compiling with YASM](https://github.com/frankplow/FFmpeg/actions/runs/4707830898/jobs/8349932081).

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -12,14 +12,13 @@ jobs:
   ffvvc-test:
     name: ffvvc-test / ${{ matrix.os.name }}/${{ matrix.compiler.name }}/${{ matrix.assembler.name }}
     env:
-      runner_threads: 4
       configure_flags: --enable-ffmpeg --disable-everything --enable-decoder=vvc --enable-parser=vvc --enable-demuxer=vvc --enable-protocol=file,pipe --enable-encoder=rawvideo --enable-muxer=rawvideo,md5
     strategy:
       fail-fast: false
       matrix:
         os: 
-          - { name: linux, runner: ubuntu-latest, shell: bash }
-          - { name: windows, runner: windows-latest, shell: 'msys2 {0}' }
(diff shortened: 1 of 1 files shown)


frankplow:
[test / windows/msvc/none failure](https://github.com/frankplow/FFmpeg/actions/runs/4707830898/jobs/8349932022) and [test / windows/msvc/nasm failure](https://github.com/ffvvc/FFmpeg/actions/runs/4707893563/jobs/8350023724) are related to #26 I believe.

nuomi2021:
> The new jobs have flagged an [error when compiling with YASM]

thank you for the pr,it has avx512 :) https://github.com/ffvvc/FFmpeg/actions/runs/4711935904/jobs/8356796175?pr=65#step:6:72
yasm issue fixed by  fixed by https://github.com/ffvvc/FFmpeg/pull/65

> If we still want to check FFVS-Project-Generator can generate a valid VS solution in CI I can add this as a separate job.

Not needed since you already use cl.exe to compile.

nuomi2021:
how about we change none to c_only or no_asm. none can mean many things, we may forget the mean in the short time.

frankplow:
Histories are behaving a little strangely because #67 was branched from [frankplow:ci-enhancements](https://github.com/frankplow/FFmpeg/tree/ci-enhancements) but merged into [FFmpeg:main](https://github.com/ffvvc/FFmpeg/tree/main) before this PR - any chance this can be merged?

frankplow:
Rebased onto main

Issue #68, opened 2023-04-24
Intermittent Windows build failure


frankplow:
Windows CI build seems to fail randomly with `unable to find mspdbcore.dll`. [Here is an example](https://github.com/ffvvc/FFmpeg/actions/runs/4787027092/jobs/8511712666). Thought this had been fixed by 68c056f071a543c50395515275a6f26e3223df64 but seemingly not.

nuomi2021:
hmm, wried

frankplow:
Potentially fixed by #70. The error is intermittent therefore I am going to leave this issue open for the time-being. Once the CI workflow has run a few more times and the error hasn't cropped up I will close this.

nuomi2021:
@frankplow ,  is it related to memory?
maybe we need limited the build thread number to 4
https://github.com/ffvvc/FFmpeg/actions/runs/4906483018/jobs/8760960018?pr=76#step:7:283

frankplow:
We haven't had a Windows build failure since 492964d2918a1cc7c2b0d8cc4e17497810152035, so I'm going to close this issue.

PR #70, opened 2023-04-29
fix negtive tests and merge check asm

nuomi2021:


nuomi2021:
Tried 2 times.
Seems https://github.com/ffvvc/FFmpeg/issues/68 fixed. @frankplow could you help try and help review?

thank you

frankplow:
@nuomi2021 Just spotted that #68 seems to only affect the main ffvvc-test build now and not checkasm. It being intermittent, I don't think it had occurred for the main test before 68c056f071a543c50395515275a6f26e3223df64 so I didn't think to make the main test build single-threaded also. Sod's law, as soon as we apply that commit it starts affecting both. I've made #71 to do the build in ffvvc-test with a single thread also. Seems to do the trick on my fork, although I've never been able to reproduce the error on there reliably for some reason.

If #71 doesn't seem to do the trick, this PR LGTM. Must admit I don't really understand why - the environment setup is identical between the ffvvc-test and checkasm jobs so combining them shouldn't change much, but if it works it works.

Sorry for not looking into this myself sooner, been busy with work.

nuomi2021:
@frankplow , if we make the main build a single thread. it may be too slow. 
The asm only take 4s better combine it with current test.

frankplow:
@nuomi2021 Yeah, #71 looks to be a few minutes slower than this PR although it does appear to fix the issue. Any chance we can run workflow dispatch a few times on this PR to confirm it fixes the issue? In this PR, FFmpeg is still being built in the same way that caused [this build error](https://github.com/ffvvc/FFmpeg/actions/runs/4787027092/jobs/8511712666#step:7:147) so I think the error might just be relatively rare.

nuomi2021:
Thank you for checking. it's really like a ffmpeg or visual studio issue. Let us merge this since it is faster. 
Thank you for you great help.

nuomi2021:
@frankplow the merge test passed [too](https://github.com/frankplow/FFmpeg/actions/runs/4838370638/jobs/8622782113). 
You are right, maybe the error is rare.

PR #71, opened 2023-04-29
CI - Build FFmpeg in ffvvc-test job using a single thread

### LINKED ISSUE #68: Intermittent Windows build failure

frankplow:
68c056f071a543c50395515275a6f26e3223df64 builds checkasm using a single thread to prevent an intermittent error `unable to find mspdbcore.dll`. This issue seems to also affect building FFmpeg in the ffvvc-test job, so this PR builds this with a single thread also.

This PR aims to close #68.


frankplow:
Made unnecessary by #70

Issue #83, opened 2023-05-30
SPS Range Extension Support


frankplow:
When using bit depths higher than 10-bit, the `sps_range_extension` may be include in the SPS. This adds five new flags, primarily to increase the dynamic range. It is possible to use higher bit depths without the range extension, however their benefit will be reduced.

frankplow:
Here are the range extension features used by the 12-bit VVCv2 bitstreams:

|Test                     |extended_precision_flag|ts_residual_coding_rice_present_in_sh_flag|rrc_rice_extension_flag|persistent_rice_adaptation_enabled_flag|reverse_last_sig_coeff_enabled_flag|
|-------------------------|-----------------------|------------------------------------------|-----------------------|---------------------------------------|-----------------------------------|
|12b400P12_A_Sony_2       |✅                      |✅                                         |✅                      |✅                                      |✅                                  |
|12b400P12_B_Sony_2       |✅                      |✅                                         |✅                      |✅                                      |✅                                  |
|12b400P12_C_Sony_2       |✅                      |✅                                         |✅                      |✅                                      |✅                                  |
|12b400P12_D_Sony_2       |✅                      |✅                                         |✅                      |✅                                      |✅                                  |
|12b400P12_E_Sony_2       |✅                      |✅                                         |✅                      |✅                                      |✅                                  |
|12b400P16_A_Sony_2       |✅                      |✅                                         |✅                      |✅                                      |✅                                  |
|12b400P16_B_Sony_2       |✅                      |✅                                         |✅                      |✅                                      |✅                                  |
|12b400P16_C_Sony_2       |✅                      |✅                                         |✅                      |✅                                      |✅                                  |
|12b400P16_D_Sony_2       |✅                      |✅                                         |✅                      |✅                                      |✅                                  |
|12b400P16_E_Sony_2       |✅                      |✅                                         |✅                      |✅                                      |✅                                  |
|12b420P12_A_Sony_2       |✅                      |✅                                         |✅                      |✅                                      |✅                                  |
|12b420P12_B_Sony_2       |✅                      |✅                                         |✅                      |✅                                      |✅                                  |
|12b420P12_C_Sony_2       |✅                      |✅                                         |✅                      |✅                                      |✅                                  |
|12b420P12_D_Sony_2       |✅                      |✅                                         |✅                      |✅                                      |✅                                  |
|12b420P12_E_Sony_2       |✅                      |✅                                         |✅                      |✅                                      |✅                                  |
[...]

PR #84, opened 2023-05-30
12-Bit Support

frankplow:
This PR adds support for 12-bit colour.

This is a draft as the tests are currently failing. Below shows a still from 12b444vvc1_B_Sony_2. There appears to be some coloured artefacts on the centre and left of the screen. Without wishing to jump to conclusions, I suspect this is an issue with the inter-prediction code.

Screenshot 2023-05-30 at 10 57 19

When using a bit depth higher than 10, the standard allows a bitstream to include the range extension in the SPS. In practice, most do this as this is required to enable the `sps_extended_precision_flag`, which increases dynamic range. There are, however, a handful of conformance bitstreams which use 12-bit colour without the range extension:
* 12b420Ivvc1_A_InterDigital_2
* 12b420SPvvc1_A_KDDI_2
* 12b420vvc1_A_Alibaba_2
* 12b444Ivvc1_A_Alibaba_2
* 12b444SPvvc1_A_Alibaba_2
* 12b444vvc1_A_Sony_2
* 12b444vvc1_B_Sony_2
* 12b444vvc1_C_Sony_2
* 12b444vvc1_D_Sony_2
* 12b444vvc1_E_Sony_2

In future, to make the most of the increased bit depths, we will likely want to implement the range extension, however this will require more substantial changes to the transform code and so is left to another PR. I have created #83 to track this.

### DIFF
--- a/libavcodec/cbs_h266_syntax_template.c
+++ b/libavcodec/cbs_h266_syntax_template.c
@@ -880,7 +880,7 @@ static int FUNC(vps)(CodedBitstreamContext *ctx, RWContext *rw,
             ues(vps_ols_dpb_pic_width[i], 0, UINT16_MAX, 1, i);
             ues(vps_ols_dpb_pic_height[i], 0, UINT16_MAX, 1, i);
             ubs(2, vps_ols_dpb_chroma_format[i], 1, i);
-            ues(vps_ols_dpb_bitdepth_minus8[i], 0, 2, 1, i);
+            ues(vps_ols_dpb_bitdepth_minus8[i], 0, 8, 1, i);
             if (vps_num_dpb_params > 1 && vps_num_dpb_params != num_multi_layer_olss)
                 ues(vps_ols_dpb_params_idx[i], 0, vps_num_dpb_params - 1, 1, i);
             else if (vps_num_dpb_params == 1)
@@ -1125,7 +1125,7 @@ static int FUNC(sps)(CodedBitstreamContext *ctx, RWContext *rw,
     }
 
 
-    ue(sps_bitdepth_minus8,   0, 2);
+    ue(sps_bitdepth_minus8, 0, 8);
     qp_bd_offset = 6 * current->sps_bitdepth_minus8;
 
     flag(sps_entropy_coding_sync_enabled_flag);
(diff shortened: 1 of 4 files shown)


frankplow:
I made a silly mistake - I was checking the decoded YUV MD5 against the provided `.md5` and not the `.yuv.md5` – doh!

Checking properly, some of the non-range-extended conformance bitstreams still fail and some pass:
✅ 12b420Ivvc1_A_InterDigital_2
✅ 12b420SPvvc1_A_KDDI_2
✅ 12b420vvc1_A_Alibaba_2
✅ 12b444Ivvc1_A_Alibaba_2
✅ 12b444SPvvc1_A_Alibaba_2
❌ 12b444vvc1_A_Sony_2
❌ 12b444vvc1_B_Sony_2
❌ 12b444vvc1_C_Sony_2
❌ 12b444vvc1_D_Sony_2
❌ 12b444vvc1_E_Sony_2

libavcodec/cbs_h266_syntax_template.c
```diff
@@ -1125,7 +1125,7 @@ static int FUNC(sps)(CodedBitstreamContext *ctx, RWContext *rw,
     }
 
 
-    ue(sps_bitdepth_minus8,   0, 2);
```
nuomi2021, line 1128: please help split it to to 4 commits, 
cbs, parser, ps, and dsp. 

thank you

frankplow:
Correction to the above - I was not setting `-vsync 0` when decoding using FFmpeg and so the video was being output with additional frames and therefore the difference was exaggerated. Doing it correctly, the first frames still match while the others differ, so I think the same conclusions can be drawn. Below shows a frame decoded by FFmpeg and the (amplified) difference between that frame and the reference decoder. The large blue artefact in the centre-left of the image can be seen on the decoded image if you look closely.

temp
temp

As well as the above, another clue pointing towards inter-prediction being the problem are the tests that pass – these are almost all entirely intra-coded. The only exception to this is 12b420vvc1_A_Alibaba_2, this uses inter-prediction yet still passes the test. Could it be that only 4:4:4 colour is affected?

nuomi2021:
thank you, Frank. 

Please also help https://github.com/ffvvc/tests/pull/3#issuecomment-1573119892

Issue #116, opened 2023-08-02
./libavcodec/x86/mathops.h:125: Error: operand type mismatch for `shr'


frankplow:
A recent update to binutils has broken FFmpeg compilation: https://trac.ffmpeg.org/ticket/10405. There is a [fix in FFmpeg master](https://git.ffmpeg.org/gitweb/ffmpeg.git/commit/effadce6c756247ea8bae32dc13bb3e6f464f0eb), so we will need to merge/rebase to allow building with binutils >= 2.41.

nuomi2021:
@frankplow , thank you for reporting this. I am working on using cbs to decoder parameter set. after it, we will rebased to latest ffmepg code

nuomi2021:
@frankplow , will freeze and rename the current main branch. and use https://github.com/ffvvc/FFmpeg/pull/118 instead.

thank you

frankplow:
@nuomi2021 #118 Does the trick for me 👍

nuomi2021:
branch switched https://github.com/ffvvc/FFmpeg

PR #114, opened 2023-07-19
1D inverse type-II DCT AVX2 optimisations

frankplow:
This PR adds AVX2 optimisations for the type-II DCT. For now, these optimisations are only implemented at the 1D level.

Performance results:

| Bitstream                             | Before | After | Delta |
|---------------------------------------|--------|-------|-------|
| RitualDance_1920x1080_60_10_420_32_LD | 101.0  | 104.0 | +3.0% |
| RitualDance_1920x1080_60_10_420_37_RA | 89.0   | 90.0  | +1.1% |
| Tango2_3840x2160_60_10_420_27_LD      | 24.0   | 25.0  | +4.2% |

```
benchmarking with Linux Perf Monitoring API
nop: 86.0
checkasm: using random seed 1542246724
AVX2:
 - vvc_itx_1d.idct2 [OK]
checkasm: all 6 tests passed
vvc_inv_dct2_2_c: 16.0
vvc_inv_dct2_2_avx2: 16.2
vvc_inv_dct2_4_c: 19.7
vvc_inv_dct2_4_avx2: 17.0
vvc_inv_dct2_8_c: 39.7
vvc_inv_dct2_8_avx2: 29.2
vvc_inv_dct2_16_c: 132.7
vvc_inv_dct2_16_avx2: 54.5
vvc_inv_dct2_32_c: 379.0
vvc_inv_dct2_32_avx2: 115.7
vvc_inv_dct2_64_c: 1527.7
vvc_inv_dct2_64_avx2: 385.0
```

PR #117, opened 2023-08-03
WIP: dav1d ITX AVX2 Assembly Port

frankplow:
This PR will port dav1d's AVX2 inverse transform optimisations.

### DIFF
--- a/libavcodec/vvc/vvc_ctu.c
+++ b/libavcodec/vvc/vvc_ctu.c
@@ -268,6 +268,8 @@ static TransformBlock* add_tb(TransformUnit *tu, VVCLocalContext *lc,
     tb->ts = 0;
     tb->coeffs = lc->coeffs;
     lc->coeffs += tb_width * tb_height;
+    tb->pixels = lc->pixels;
+    lc->pixels += tb_width * tb_height;
     return tb;
 }
 
@@ -2382,6 +2384,7 @@ int ff_vvc_coding_tree_unit(VVCLocalContext *lc,
     }
 
     lc->coeffs = fc->tab.coeffs + rs * ctb_size * VVC_MAX_SAMPLE_ARRAYS;
+    lc->pixels = fc->tab.pixels + rs * ctb_size * VVC_MAX_SAMPLE_ARRAYS;
     lc->cu     = NULL;
 
     ff_vvc_cabac_init(lc, ctu_idx, rx, ry);
--- a/libavcodec/vvc/vvc_ctu.h
+++ b/libavcodec/vvc/vvc_ctu.h
@@ -96,6 +96,7 @@ typedef struct TransformBlock {
     int bd_offset;
 
     int *coeffs;
+    int16_t *pixels;
 } TransformBlock;
 
 typedef enum VVCTreeType {
@@ -369,6 +370,7 @@ struct VVCLocalContext {
     VVCFrameContext *fc;
     EntryPoint *ep;
     int *coeffs;
+    int16_t *pixels;
 } ;
 
 typedef struct VVCAllowedSplit {
--- a/libavcodec/vvc/vvc_intra.c
+++ b/libavcodec/vvc/vvc_intra.c
@@ -272,32 +272,6 @@ static void predict_intra(VVCLocalContext *lc, const TransformUnit *tu, const in
     }
 }
 
-static void scale_clip(int *coeff, const int nzw, const int w, const int h,
-    const int shift, const int log2_transform_range)
-{
-    const int add = 1 << (shift - 1);
-    for (int y = 0; y < h; y++) {
-        int *p = coeff + y * w;
-        for (int x = 0; x < nzw; x++) {
-            *p = av_clip_intp2((*p + add) >> shift, log2_transform_range);
-            p++;
-        }
-        memset(p, 0, sizeof(*p) * (w - nzw));
-    }
-}
-
-static void scale(int *out, const int *in, const int w, const int h, const int shift)
-{
-    const int add = 1 << (shift - 1);
-    for (int y = 0; y < h; y++) {
-        for (int x = 0; x < w; x++) {
-            int *o = out + y * w + x;
-            const int *i = in + y * w + x;
-            *o = (*i + add) >> shift;
-        }
-    }
-}
-
 // part of 8.7.3 Scaling process for transform coefficients
 static void derive_qp(const VVCLocalContext *lc, const TransformUnit *tu, TransformBlock *tb)
 {
@@ -441,35 +415,6 @@ static void dequant(const VVCLocalContext *lc, const TransformUnit *tu, Transfor
     }
 }
 
-static void itx_2d(const VVCFrameContext *fc, TransformBlock *tb, const enum TxType trh, const enum TxType trv, int *temp)
-{
-    const VVCSPS *sps   = fc->ps.sps;
-    const int w         = tb->tb_width;
-    const int h         = tb->tb_height;
-    const int nzw       = tb->max_scan_x + 1;
-
-    for (int x = 0; x < nzw; x++)
-        fc->vvcdsp.itx.itx[trv][tb->log2_tb_height - 1](temp + x, w, tb->coeffs + x, w);
-    scale_clip(temp, nzw, w, h, 7, sps->log2_transform_range);
-
-    for (int y = 0; y < h; y++)
-        fc->vvcdsp.itx.itx[trh][tb->log2_tb_width - 1](tb->coeffs + y * w, 1, temp + y * w, 1);
-    scale(tb->coeffs, tb->coeffs, w, h, 5 + sps->log2_transform_range - sps->bit_depth);
-}
-
-static void itx_1d(const VVCFrameContext *fc, TransformBlock *tb, const enum TxType trh, const enum TxType trv, int  *temp)
-{
-    const VVCSPS *sps   = fc->ps.sps;
-    const int w         = tb->tb_width;
-    const int h         = tb->tb_height;
-
-    if (w > 1)
-        fc->vvcdsp.itx.itx[trh][tb->log2_tb_width - 1](temp, 1, tb->coeffs, 1);
-    else
-        fc->vvcdsp.itx.itx[trv][tb->log2_tb_height - 1](temp, 1, tb->coeffs, 1);
-    scale(tb->coeffs, temp, w, h, 6 + sps->log2_transform_range - sps->bit_depth);
-}
-
 static void transform_bdpcm(TransformBlock *tb, const VVCLocalContext *lc, const CodingUnit *cu)
 {
     const VVCSPS *sps        = lc->fc->ps.sps;
@@ -511,14 +456,14 @@ static void itransform(VVCLocalContext *lc, TransformUnit *tu, const int tu_idx,
             dequant(lc, tu, tb);
             if (!tb->ts) {
                 enum TxType trh, trv;
+                const int nzw = tb->max_scan_x + 1;
 
                 if (cu->apply_lfnst_flag[c_idx])
                     ilfnst_transform(lc, tb);
                 derive_transform_type(fc, lc, tb, &trh, &trv);
-                if (w > 1 && h > 1)
-                    itx_2d(fc, tb, trh, trv, temp);
-                else
-                    itx_1d(fc, tb, trh, trv, temp);
+
+                fc->vvcdsp.itx.itx[trh][trv][tb->log2_tb_width][tb->log2_tb_height](
+                        tb->coeffs, tb->coeffs, nzw, sps->log2_transform_range);
             }
 
             if (chroma_scale)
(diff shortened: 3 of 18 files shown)


Jamaika1:
```
vvcdsp_init.c:264:33: error: assignment to expression with array type
  264 |     c->itx.itx[DCT2][0]         = ff_vvc_inv_dct2_2_##opt;                      \
      |                                 ^
vvcdsp_init.c:279:13: note: in expansion of macro 'IDCT2_INIT'
  279 |             IDCT2_INIT(avx2);
      |             ^~~~~~~~~~
```
Can add definition
#if HAVE_AVX2_EXTERNAL
...
#endif

frankplow:
@Jamaika1 
> ```
> vvcdsp_init.c:264:33: error: assignment to expression with array type
>   264 |     c->itx.itx[DCT2][0]         = ff_vvc_inv_dct2_2_##opt;                      \
>       |                                 ^
> vvcdsp_init.c:279:13: note: in expansion of macro 'IDCT2_INIT'
>   279 |             IDCT2_INIT(avx2);
>       |             ^~~~~~~~~~
> ```
> 
> Can add definition #if HAVE_AVX2_EXTERNAL ... #endif

I'm sorry I don't see where his code is from - there are currently no AVX2 optimisations implemented on this branch and it looks to me like the version of vvcdsp_init.c this is referencing is from my [frankplow:idct-asm](https://github.com/frankplow/FFmpeg/tree/idct-asm) branch.

frankplow:
Rebase and re-target onto new main, to reflect #116.

frankplow:
@nuomi2021 Not urgent at the minute, but I thought I should make a comment about this before I forget it. Can you think of a better way to deal with the memory for transform blocks' transform coefficients and pixels?

After 36e417619cc69865cf705ca02ff388fce4bf9a94 the transform coefficients and pixels are allocated separately in the `VVCFrameContext`:
https://github.com/ffvvc/FFmpeg/blob/36e417619cc69865cf705ca02ff388fce4bf9a94/libavcodec/vvc/vvcdec.c#L98-L103
However this is redundant — `pixels` is not used before the transform and `coeffs` is not used afterwards.

I wondered if it would be possible to allocate for the coefficients statically for each transform block, for example at the top of [`itransform`](https://github.com/ffvvc/FFmpeg/blob/2a7d709ad771c8cba4b14fb630b378740253ab27/libavcodec/vvc/vvc_intra.c#L431), however this is not possible as the CABAC must be able to put data into the coefficients array:
https://github.com/ffvvc/FFmpeg/blob/2a7d709ad771c8cba4b14fb630b378740253ab27/libavcodec/vvc/vvc_cabac.c#L2145-L2152

Perhaps there is some way the same memory could be used for both `pixels` and `coeffs`, as before, but reinterpreted as `int16_t`s or `int`s at different points in the program? The size of a pixel is always <= the size of a transform coefficient in the current version of VVC, the extreme case being if the bit depth is 16 and `extended_precision_flag` is 0, then both are 16 bits. Would just setting `tb->pixels = (int16_t *) tb->coeffs` be acceptable given this is the case?

frankplow:
@nuomi2021 I have spent some more time trying to port the dav1d assembly to FFVVC, and I have come to the conclusion that it will not be possible. The reasoning for this is quite nuanced and mathematical so please bear with me. I am sure you are familiar with a good deal of the background here but I have added some for quick reference and to aid any other readers.

Any N-point 1D DCT (-II) can be expressed as a matrix multiplication, requiring N² multiplications and N² - N additions. This is not the most efficient way to compute the DCT however — the elements of the transform matrix are values of the cosine function, and so are not independent. In particular, the matrix exhibits various types of redundancies. This includes simple symmetries which can be accounted for using butterfly algorithms, as well as more complex relationships which rely on trigonometric identities. Fast DCT algorithms leverage these redundancies to compute the DCT with fewer multiplications and additions. dav1d in particular uses the Chen algorithm [1] to compute the DCT. A more recent description of this algorithm can be found in [2].

Modern video coding standards do not use the actual DCT, whose transform coefficients are real numbers between -1 and 1, but instead an approximation of it using only integers. This is done for a number of pragmatic reasons centring around efficiency and consistency. Simply multiplying the DCT by a constant and rounding to the nearest integer produces an accurate approximation of the DCT and therefore a good compression performance, however destroys many of the redundancies in the DCT. The design of an effective DCT approximation is a bit of an art form involving the balancing of these two competing factors: accuracy to the DCT, which results in good compression, and preservation of redundancies, which results in good performance.

Unfortunately, the approximation used in HEVC and the derived approximation in VVC, do not preserve the trigonometric redundancies required to employ the Chen algorithm used in the dav1d transform optimisation [3, section 6.2.2]. This makes the dav1d ITX assembly unusable for VVC unfortunately.

I should have caught this earlier really, but my plan now is to revise my PR #114 to use the work done in  https://github.com/ffvvc/FFmpeg/pull/117/commits/ea7a0ceec6041c12af370f6056cdf2939e0f82f5 and https://github.com/ffvvc/FFmpeg/pull/117/commits/36e417619cc69865cf705ca02ff388fce4bf9a94, as well as some other ideas I have seen in my work with the dav1d assembly. I hope I can make these optimisations several times faster than they already are, but fundamentally the speed is going to be limited by the algorithm that can be used.

[1] Wen-Hsiung Chen, C. Smith, S.Fralick, A Fast Computational Algorithm for the Discrete Cosine Transform, [DOI 10.1109/TCOM.1977.1093941](https://doi.org/10.1109/TCOM.1977.1093941), ([free pdf](https://www.cin.ufpe.br/~vak/tg/Chen-dct.pdf))
[2] C. J. Tablada, T. L. T. da Silveira, R. J. Cintra, F. M. Bayer, DCT Approximations Based on Chen's Factorization, [DOI 10.1016/j.image.2017.06.014](https://doi.org/10.1016/j.image.2017.06.014), ([free preprint](https://arxiv.org/abs/2207.11638))
[3] Vivienne Sze, Madhukar Budagavi, Gary J. Sullivan, High Efficiency Video Coding (HEVC) Algorithms and Architectures, [DOI 10.1007/978-3-319-06895-4](https://doi.org/10.1007/978-3-319-06895-4)

nuomi2021:
> @nuomi2021 Not urgent at the minute, but I thought I should make a comment about this before I forget it. Can you think of a better way to deal with the memory for transform blocks' transform coefficients and pixels?
> 
> After [36e4176](https://github.com/ffvvc/FFmpeg/commit/36e417619cc69865cf705ca02ff388fce4bf9a94) the transform coefficients and pixels are allocated separately in the `VVCFrameContext`:
> 
> https://github.com/ffvvc/FFmpeg/blob/36e417619cc69865cf705ca02ff388fce4bf9a94/libavcodec/vvc/vvcdec.c#L98-L103
> 
> 
> However this is redundant — `pixels` is not used before the transform and `coeffs` is not used afterwards.
Not possible. coeffs has a different layout than pixels. Suppose you have 4 64x64 blocks. each block has 1 coeffs. 
We only use the first 4 coeffs. combine coeffs buffer with pixels will overite the second blocks coeffs

> I wondered if it would be possible to allocate for the coefficients statically for each transform block, for example at the top of [`itransform`](https://github.com/ffvvc/FFmpeg/blob/2a7d709ad771c8cba4b14fb630b378740253ab27/libavcodec/vvc/vvc_intra.c#L431), however this is not possible as the CABAC must be able to put data into the coefficients array:
> 
> https://github.com/ffvvc/FFmpeg/blob/2a7d709ad771c8cba4b14fb630b378740253ab27/libavcodec/vvc/vvc_cabac.c#L2145-L2152
> 
> Perhaps there is some way the same memory could be used for both `pixels` and `coeffs`, as before, but reinterpreted as `int16_t`s or `int`s at different points in the program? The size of a pixel is always <= the size of a transform coefficient in the current version of VVC, the extreme case being if the bit depth is 16 and `extended_precision_flag` is 0, then both are 16 bits. Would just setting `tb->pixels = (int16_t *) tb->coeffs` be acceptable given this is the case?

Could you explain why you want to do this?
thank you

nuomi2021:
> @nuomi2021 I have spent some more time trying to port the dav1d assembly to FFVVC, and I have come to the conclusion that it will not be possible. The reasoning for this is quite nuanced and mathematical so please bear with me. I am sure you are familiar with a good deal of the background here but I have added some for quick reference and to aid any other readers.
> 
> Any N-point 1D DCT (-II) can be expressed as a matrix multiplication, requiring N² multiplications and N² - N additions. This is not the most efficient way to compute the DCT however — the elements of the transform matrix are values of the cosine function, and so are not independent. In particular, the matrix exhibits various types of redundancies. This includes simple symmetries which can be accounted for using butterfly algorithms, as well as more complex relationships which rely on trigonometric identities. Fast DCT algorithms leverage these redundancies to compute the DCT with fewer multiplications and additions. dav1d in particular uses the Chen algorithm [1] to compute the DCT. A more recent description of this algorithm can be found in [2].
> 
> Modern video coding standards do not use the actual DCT, whose transform coefficients are real numbers between -1 and 1, but instead an approximation of it using only integers. This is done for a number of pragmatic reasons centring around efficiency and consistency. Simply multiplying the DCT by a constant and rounding to the nearest integer produces an accurate approximation of the DCT and therefore a good compression performance, however destroys many of the redundancies in the DCT. The design of an effective DCT approximation is a bit of an art form involving the balancing of these two competing factors: accuracy to the DCT, which results in good compression, and preservation of redundancies, which results in good performance.
> 
> Unfortunately, the approximation used in HEVC and the derived approximation in VVC, do not preserve the trigonometric redundancies required to employ the Chen algorithm used in the dav1d transform optimisation [3, section 6.2.2]. This makes the dav1d ITX assembly unusable for VVC.
> 
> I should have caught this earlier really, but my plan now is to revise my PR #114 to use the work done in [ea7a0ce](https://github.com/ffvvc/FFmpeg/commit/ea7a0ceec6041c12af370f6056cdf2939e0f82f5) and [36e4176](https://github.com/ffvvc/FFmpeg/commit/36e417619cc69865cf705ca02ff388fce4bf9a94), as well as some other ideas I have seen in my work with the dav1d assembly. I hope I can make these optimisations several times faster than they already are, but fundamentally the speed is going to be limited by the slower Hung algorithm that is supported by VVC
Thank you for your hardwork.

No worries. Please continue working on https://github.com/ffvvc/FFmpeg/pull/114
In the meaning time, could you check vvdec c and asm code?
> 
> [1] Wen-Hsiung Chen, C. Smith, S.Fralick, A Fast Computational Algorithm for the Discrete Cosine Transform, [DOI 10.1109/TCOM.1977.1093941](https://doi.org/10.1109/TCOM.1977.1093941), ([free pdf](https://www.cin.ufpe.br/~vak/tg/Chen-dct.pdf)) [2] C. J. Tablada, T. L. T. da Silveira, R. J. Cintra, F. M. Bayer, DCT Approximations Based on Chen's Factorization, [DOI 10.1016/j.image.2017.06.014](https://doi.org/10.1016/j.image.2017.06.014), ([free preprint](https://arxiv.org/abs/2207.11638)) [3] Vivienne Sze, Madhukar Budagavi, Gary J. Sullivan, High Efficiency Video Coding (HEVC) Algorithms and Architectures, [DOI 10.1007/978-3-319-06895-4](https://doi.org/10.1007/978-3-319-06895-4)

frankplow:
@nuomi2021 
> No worries. Please continue working on #114 In the meaning time, could you check vvdec c and asm code?

vvdec has a very straightforward implementation in assembly. It does not use any fast algorithm like Hung, instead the transform is a simple matrix multiplication. I have read that in some instances this may be faster as it is more readily parallelisable, but I am not convinced — using `perf.py` I measure only a 1.1% performance improvement when toggling vvdec's transform optimisations, which is less than the difference I got with my optimisations using the Hung algorithm in #114. Additionally, it would be quite a lot of work to port these, as they are written using intrinsics and C++ templates.

I've also looked into adapting the assembly optimisations already in FFmpeg for HEVC. These are written directly in NASM assembly using x86inc.asm, so would require less work to port on that front. Additionally, they use the Hung algorithm, which theoretically should speed them up. The big obstacle in adapting them is the fact that they use 16-bit integers for both pixels and transform coefficients. This is not suitable for VVC as when `sps_extended_precision_flag` is set, the transform coefficients are `bit_depth + 6` and so can be up to 22-bit. I can see two ways around this:
1. The easiest fix is to convert the 32-bit transform coefficients to 16-bit within the functions when possible, and disable the optimisations otherwise. This is complicated by the fact that these optimisations do a lot of writing to and from memory, and so there is no one easy place where `mova`s to load from memory can be replaced with `packssdw`s for example. I have already implemented the naive approach of doing O(n²) `packssdw`s at the beginning of the function, however this degraded the performance to below the C implementation.
2. Alternatively, the function prototype itself could be changed to use 16-bit transform coefficients, either when the bit depth <= 10 (this simplies `sps_extended_precision_flag = 0`) or based on `sps_extended_precision_flag` itself. This wouldn't require major changes to the assembly implementation, however there is currently no mechanism to have function prototypes change based on bit depth. I think this is the more worthwhile solution, as the ability to use smaller types when possible enables faster optimisations, but how much work do you think this would be?

In either case, the 2x2, 64x64, 1D and rectangular transforms would also need to be implemented as these are not present in HEVC.

Issue #119, opened 2023-08-13
sao asm mismatch with hevc


nuomi2021:
There are several code mismatches with hevc. 
like https://github.com/ffvvc/FFmpeg/blob/main/libavcodec/x86/vvc_sao.asm#L193, the EDGE_SRCSTRIDE 
 and 
https://github.com/ffvvc/FFmpeg/blob/main/libavcodec/x86/vvc_sao.asm#L173
@zackerthescar do you know why?

thank you

zackerthescar:
> https://github.com/ffvvc/FFmpeg/blob/main/libavcodec/x86/vvc_sao.asm#L193

I had found out that VVC's edge stride was double that of HEVC's, but I didn't realize that this is because of the PB size. Fixed it in #121 

> https://github.com/ffvvc/FFmpeg/blob/main/libavcodec/x86/vvc_sao.asm#L173

HEVC SAO uses only SSSE3 functions for the 8 and 16 case regardless of if the CPU supports AVX2. Since we're targeting AVX2, we have to declare these functions as AVX2 instead.

Declaring `VVC_SAO_EDGE_FILTER 8,0,u` after `INIT_YMM` results in compilation errors, because of the `if` codeblock on 281-287.

nuomi2021:
Any machine support avx2 will have ssse3. 
Due to data length reasons. use avx2 may not get benefits. It's not a fault to use ssse3 in this case

PR #121, opened 2023-08-13
vvc_sao: proper fix for doubled PB size constant

zackerthescar:
_Actually_ fixes the doubled PB size constant. Stupid error on my end, to be honest.

Fixes the first part of #119


nuomi2021:
No worries. 
Compare to hevc, vvc's ctu size is 128. so we change MAX_PB_SIZE TO 128.

PR #130, opened 2023-08-27
FFmpeg HEVC IDCT port

frankplow:
This PR ports FFmpeg's HEVC IDCT optimisations.

To-do:
- [x] Port 4x4 transform
- [x] Port 8x8 transform
- [x] Port 16x16 transform
- [ ] Port 32x32 transform
- [ ] Add 2x2 transform
- [ ] Add 64x64 transform
- [ ] Add rectangular transforms
- [ ] Add 1D transforms

Checkasm benchmark results:
```
inv_dct2_dct2_4x4_8_c: 226.2
inv_dct2_dct2_4x4_8_avx2: 26.2
inv_dct2_dct2_4x4_10_c: 188.2
inv_dct2_dct2_4x4_10_avx2: 24.7
inv_dct2_dct2_8x8_8_c: 704.7
inv_dct2_dct2_8x8_8_avx2: 124.7
inv_dct2_dct2_8x8_10_c: 751.2
inv_dct2_dct2_8x8_10_avx2: 124.7
inv_dct2_dct2_16x16_8_c: 4289.7
inv_dct2_dct2_16x16_8_avx2: 621.2
inv_dct2_dct2_16x16_10_c: 4335.2
inv_dct2_dct2_16x16_10_avx2: 625.2
```

`perf.py` results:
| Bitstream                             | Before | After | Delta |
|---------------------------------------|--------|-------|-------|
| RitualDance_1920x1080_60_10_420_32_LD | 99.7   | 99.3  | -0.4% |
| RitualDance_1920x1080_60_10_420_37_RA | 88.3   | 87.7  | -0.5% |
| Tango2_3840x2160_60_10_420_27_LD      | 23.0   | 23.0  |  0.0% |

The current `perf.py` performance is poor as the DCT's effect on overall decoding performance is dominated by the larger sizes which have not yet been implemented. The decrease in performance is explained by the additional overhead of optimising at the 2D level, the benefits of which are not being reaped here. As the larger sizes are implemented, performance will increase dramatically, in line with the checkasm benchmark result.


libavcodec/vvc/vvc_ctu.h
```diff
@@ -96,6 +96,7 @@ typedef struct TransformBlock {
     int bd_offset;
 
     int *coeffs;
+    int16_t *pixels;
```
nuomi2021, line 99: why do we need pixels? can we just change coeffs to int16_t?

frankplow: If `sps_extended_precision_flag = 1`, then the transform coefficients are `bit_depth + 6` bits each.

frankplow: Note that if `bit_depth <= 10`, `sps_extended_precision_flag` must be 0. For now, I think it is easiest to treat transform coeffs as capable of fitting in a `int16_t` for `bit_depth <= 10` and not for `bit_depth > 10`, as we already have the mechanism to use different functions depending on bit-depth. This is what the current code is already doing. In the long term, yes I agree it will be better to change based on `sps_extended_precision_flag`, but as there are few > 10-bit videos which do not use the range extension I don't think it needs to be a priority.

Re. `coeffs`/`pixels`, this is the same problem I was trying to get at in https://github.com/ffvvc/FFmpeg/pull/117#issuecomment-1682161165. I'm not quite sure why you suggest `uint8_t *` — is this just meant as an arbitrary type which will be reinterpreted according to when an `int` or `int16_t` is needed? If that's the case I think `char` may be more fitting.

Aside: I nearly have the 32x32 transform working, there is just a single conformance test failing due to an overflow from a very large transform coefficient.

nuomi2021: > Re. coeffs/pixels, this is the same problem I was trying to get at in https://github.com/ffvvc/FFmpeg/pull/117#issuecomment-1682161165. I'm not quite sure why you suggest uint8_t * — is this just meant as an arbitrary type which will be reinterpreted according to when an int or int16_t is needed? If that's the case I think char may be more fitting.

sizeof(char) is not always 8, according to spec. it only can make sure sizeof(char) is >= 8. https://en.cppreference.com/w/cpp/language/types

we can define a new type coeff like this

```
#if BIT_DEPTH > 10 && EXTEND_PRECISION
 typedef Coeff int32_t;
#define TR_FUNC(f) FUNC(f##_ep)
#else 
 typedef Coeff int16_t;
#define TR_FUNC(f) FUNC(f)
#endif

static void TR_FUNC(transform_bdpcm)(uint8_t *_coeffs, const int width, const int height,
    const int vertical, const int log2_transform_range)
{
          Coeff *coeff = (Coeff*)_coeffs
...
}
```

frankplow: Here's something relevant I've found while getting 88f8a7dd21b91c20a9db28fe5c4788a9d05de4ce working. The residual samples produced by the transform process may be bigger than the eventual reconstructed pixels. For example, the 10-bit 32x32 DCT-II may produce residuals which are greater than `(1 << 10 - 1) - 1` and even greater than `INT16_MAX`. The value is only clipped _after_ being summed with the predicted sample. I think we may then need to use a larger type for the output of the transforms. I am not sure quite what the bound is on the residual samples, but I think `int32_t`s should be sufficient.

At the moment, this is manifesting in an intermittent failure for the ITX checkasm of `ff_vvc_inv_dct2_dct2_32x32_10_avx2`. In the case the output is so large it overflows the `int16_t` pixel type, the C implementation wraps around whereas the assembly implementation saturates, so the test fails. The patch below fails reliably.

<details>
<summary><b>Patch for tests/checkasm/vvc_itx.c to set all transform coefficients to maximum value</b></summary>

```diff
diff --git a/tests/checkasm/vvc_itx.c b/tests/checkasm/vvc_itx.c
index f403c4f1ac..17b18bcc03 100644
--- a/tests/checkasm/vvc_itx.c
+++ b/tests/checkasm/vvc_itx.c
@@ -35,7 +35,7 @@
         int k;                                                                  \
         for (k = 0; k < size; ++k) {                                            \
             uint32_t r = rnd();                                                 \
-            int32_t a = min + r / (UINT32_MAX / (max - min + 1) + 1);           \
+            int32_t a = min + UINT32_MAX / (UINT32_MAX / (max - min + 1) + 1);  \
             AV_WN32A(buf0 + k, a);                                              \
             AV_WN32A(buf1 + k, a);                                              \
         }                                                                       \
```

</details>

nuomi2021: > Rather than reinterpreting the uint8_t parameter, can we not just use Coeff in the signature? Would need some extra work in vvcdsp.h but probably beneficial in the long run.
Not possible. Coeff will defined to uint16_t and uint32_t in condition. we can't not use it as a signature.

thank you

nuomi2021: >The value is only clipped after being summed with the predicted sample
If it's only internal overflow(in the function), maybe you can use int32_t as an internal type. 

>At the moment, this is manifesting in an intermittent failure for the ITX checkasm of ff_vvc_inv_dct2_dct2_32x32_10_avx2
why overflow only happens at 32x32? will  16x16 .. 2x2 has the same issue? will 8 bits has the same issue?

frankplow: > > Rather than reinterpreting the uint8_t parameter, can we not just use Coeff in the signature? Would need some extra work in vvcdsp.h but probably beneficial in the long run.
> 
> Not possible. Coeff will defined to uint16_t and uint32_t in condition. we can't not use it as a signature.
> 
> thank you

Ah I see. dav1d uses their coeff/pixel types in their signatures, and I thought this was a template which was instantiated for each bit depth, but it looks as though it is only a build-time selection.

I am still not sure about the `uint8_t *` type though — if you were to do any manipulations on `coeffs` as `uint8_t`s it would cause corruption. Maybe a `void *` to make it more explicit that the array will need to be interpreted according to the bit depth and `sps_extended_precision_flag`?

frankplow: > > The value is only clipped after being summed with the predicted sample
>
> If it's only internal overflow(in the function), maybe you can use int32_t as an internal type.

Unfortunately this summing does not take place in the transforms but instead later in `add_residual`. Integrating it into the transforms would be difficult as there are many different ways the residual samples are produced before this, e.g. using core transforms, LFNST, transform skip, etc.

> > At the moment, this is manifesting in an intermittent failure for the ITX checkasm of ff_vvc_inv_dct2_dct2_32x32_10_avx2
>
> why overflow only happens at 32x32? will  16x16 .. 2x2 has the same issue? will 8 bits has the same issue?

Larger transforms produce larger residual samples simply because there are more terms in the dot products making up the matrix multiplication. The 8-bit 32x32 transform is not affected as its `bd_shift` is larger, enough to guarantee the samples remain below `INT16_MAX`. I have not yet implemented the 10-bit 64x64 transform, though I am fairly certain it will suffer from the same issue. I also think the non-extended-precision 12-bit+ 32x32 and 64x64 transforms are affected. There is also the possibility that the 8-bit 64x64 transform is affected and that smaller non-extended-precision 12-bit+ transforms are affected.

This is not really an issue in any sensible video — why would the residual ever be many times larger than the full range of a pixel? — hence not being caught by any of the conformance tests.

nuomi2021: > Ah I see. dav1d uses their coeff/pixel types in their signatures, and I thought this was a template which was instantiated for each bit depth, but it looks as though it is only a build-time selection.

the dav1d uses templates at a higher level. all bytefn(fn) are template functions.

>The 8-bit 32x32 transform is not affected as its bd_shift is larger, enough to guarantee the samples remain below INT16_MAX
If you check this line https://github.com/ffvvc/FFmpeg/blob/main/libavcodec/vvc/vvc_intra.c#L470
It means the input depth is 5 + sps->log2_transform_range, the output depth is sps->bit_depth. 
So the itx's output depth should be 5 + sps->log2_transform_range, not related to bit depth.
the range of sps->log_transform_range is [15, 20], so 5 + sps->log2_transform_range is [20, 25].
we can't hold it in int16_t. we need int32_t in any case.

libavcodec/vvc/vvcdsp_template.c
```diff
@@ -93,26 +94,164 @@ static void FUNC(transform_bdpcm)(int *coeffs, const int width, const int height
...
+    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 32, 16);                             \
+    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 32, 32);
 
-#define VVC_ITX_COMMON(TYPE, type)                                              \
-        VVC_ITX(TYPE, type, 4);                                                 \
-        VVC_ITX(TYPE, type, 8);                                                 \
-        VVC_ITX(TYPE, type, 16);                                                \
-        VVC_ITX(TYPE, type, 32);
+#define ITX \
+    ITX_COMMON_SIZES(DCT2, dct2, DCT2, dct2);                                   \
```
nuomi2021, line 124: maybe you can define 
```
#define ITX_2D_WX(TYPE_H, type_h, TYPE_V, type_v, W) \
    ITX_2D_(TYPE_H, type_h, TYPE_V, type_v, W, 2); \
    ITX_2D_(TYPE_H, type_h, TYPE_V, type_v, W, 4）; \
    ..
```

then you can do 

```
#define ITX_COMMON_SIZES(DCT2, dct2, DCT2, dct2) \
    ITX_2D_WX(DCT2, dct2, DCT2, dct2, 2);
    ....

```

libavcodec/vvc/vvcdsp.h
```diff
@@ -111,11 +112,10 @@ typedef struct VVCIntraDSPContext {
...
 typedef struct VVCItxDSPContext {
-    void (*add_residual)(uint8_t *dst, const int *res, int width, int height, ptrdiff_t stride);
-    void (*add_residual_joint)(uint8_t *dst, const int *res, int width, int height, ptrdiff_t stride, int c_sign, int shift);
-    void (*pred_residual_joint)(int *buf, int width, int height, int c_sign, int shift);
-
-    void (*itx[N_TX_TYPE][N_TX_SIZE])(int *out, ptrdiff_t out_step, const int *in, ptrdiff_t in_step);
+    void (*add_residual)(uint8_t *dst, const int16_t *res, int width, int height, ptrdiff_t stride);
+    void (*add_residual_joint)(uint8_t *dst, const int16_t *res, int width, int height, ptrdiff_t stride, int c_sign, int shift);
+    void (*pred_residual_joint)(int16_t *buf, int width, int height, int c_sign, int shift);
+    void (*itx[N_TX_TYPE][N_TX_TYPE][N_TX_SIZE][N_TX_SIZE])(int16_t *dst, const int *coeff, int nzw, int log2_transform_range);
```
nuomi2021, line 118: DST7 64 and DST7 2 is not existed. 
this will make the function include a a NULL pointer. 

Could we define like this:
```
Enum TxType {
   DCT2_1,
   DCT2_2,
   ....
   DCT2_64,
   DST7_1
   DST7_4
   ...
   DST7_32
};
```

then we can use define itx as itx[TxType][TxType]

libavcodec/vvc/vvcdsp_template.c
frankplow: Like this:
```diff
diff --git a/libavcodec/vvc/vvcdsp_template.c b/libavcodec/vvc/vvcdsp_template.c
index 4cba360411..cf9fd2f69e 100644
--- a/libavcodec/vvc/vvcdsp_template.c
+++ b/libavcodec/vvc/vvcdsp_template.c
@@ -94,31 +94,19 @@ static void FUNC(transform_bdpcm)(int *coeffs, const int width, const int height
     }
 }
 
+#define ITX_COMMON_WIDTHS(TYPE_H, type_h, TYPE_V, type_v, height)               \
+    ITX_1D_V(TYPE_H, type_h, TYPE_V, type_v, 1, height)                         \
+    ITX_1D_V(TYPE_H, type_h, TYPE_V, type_v, 4, height)                         \
+    ITX_1D_V(TYPE_H, type_h, TYPE_V, type_v, 8, height)                         \
+    ITX_1D_V(TYPE_H, type_h, TYPE_V, type_v, 16, height)                        \
+    ITX_1D_V(TYPE_H, type_h, TYPE_V, type_v, 32, height)                        \
+
 #define ITX_COMMON_SIZES(TYPE_H, type_h, TYPE_V, type_v)                        \
-    ITX_1D_V(TYPE_H, type_h, TYPE_V, type_v, 1, 4);                             \
-    ITX_1D_V(TYPE_H, type_h, TYPE_V, type_v, 1, 8);                             \
-    ITX_1D_V(TYPE_H, type_h, TYPE_V, type_v, 1, 16);                            \
-    ITX_1D_V(TYPE_H, type_h, TYPE_V, type_v, 1, 32);                            \
-    ITX_1D_H(TYPE_H, type_h, TYPE_V, type_v, 4, 1);                             \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 4, 4);                               \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 4, 8);                               \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 4, 16);                              \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 4, 32);                              \
-    ITX_1D_H(TYPE_H, type_h, TYPE_V, type_v, 8, 1);                             \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 8, 4);                               \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 8, 8);                               \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 8, 16);                              \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 8, 32);                              \
-    ITX_1D_H(TYPE_H, type_h, TYPE_V, type_v, 16, 1);                            \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 16, 4);                              \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 16, 8);                              \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 16, 16);                             \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 16, 32);                             \
-    ITX_1D_H(TYPE_H, type_h, TYPE_V, type_v, 32, 1);                            \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 32, 4);                              \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 32, 8);                              \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 32, 16);                             \
-    ITX_2D(TYPE_H, type_h, TYPE_V, type_v, 32, 32);
+    ITX_COMMON_WIDTHS(TYPE_h, type_h, TYPE_V, type_v, 1)                        \
+    ITX_COMMON_WIDTHS(TYPE_h, type_h, TYPE_V, type_v, 4)                        \
+    ITX_COMMON_WIDTHS(TYPE_h, type_h, TYPE_V, type_v, 8)                        \
+    ITX_COMMON_WIDTHS(TYPE_h, type_h, TYPE_V, type_v, 16)                       \
+    ITX_COMMON_WIDTHS(TYPE_h, type_h, TYPE_V, type_v, 32)
 
 #define ITX \
     ITX_COMMON_SIZES(DCT2, dct2, DCT2, dct2);                                   \
```
do you mean?

This is slightly less readable imo and it only saves 12 lines, but I am happy to add it if you think it's better. 

Sidenote: the patch above actually fails to compile at the moment as I have not implemented the size-1 1-D transforms (i.e. for transform size 1x1). These an edge case and are not used in any of the conformance bitstreams, but I think they may actually be valid per the spec and should be implemented.

libavcodec/vvc/vvcdsp.h
frankplow: Implemented in 140554f086901a0a898ea3daefb16e7c37e5ccee.

frankplow:
> hi @frankplow , seems the int32_t is only needed by range extension. If range extension is not enabled, we can keep the transform coeffs as int16_t. I will try to make some changes to this. Hope this will reduce the porting efforts.

Yeah I think if we take this approach, it shouldn't be too hard to get transforms implemented for the square sizes. Unfortunately, I think it will be hard to extend the HEVC optimisations to rectangular sizes and MTS as the way it's written doesn't facilitate much code reuse/modularity. I have a branch where I've worked on a more modular optimisation, based on some of the custom ABI ideas dav1d uses, but I'm having to write this from the ground up and don't have much time alongside my Master's at the moment. I think then, the best way to get optimisations in for these most common square sizes is to, as you say, allow varying the coeff type based on whether the range extension is active and then port the HEVC transforms.

nuomi2021:
> but I'm having to write this from the ground up and don't have much time alongside my Master's at the moment

No worries. I will continue your work after I have done the thread optimizations.

PR #140, opened 2023-09-07
transform: change coeffs type from int to int32_t

nuomi2021:


nuomi2021:
@frankplow . please help review. thank you

frankplow:
All LGTM, although I think it will be easier to implement this change as a commit on the #130 branch.

nuomi2021:
sure, let me close this. 
could help create a commit to #130? thank you

frankplow:
👍 I will get this done once I have the size-64 and rectangular transforms working. Note that it will require a significant refactor of the assembly optimisations.

PR #184, opened 2024-01-15
reduce memset bytes for frame init

nuomi2021:
we init them in the main thread, it will become bottleneck when resolution is high

### DIFF
--- a/libavcodec/vvc/vvcdec.c
+++ b/libavcodec/vvc/vvcdec.c
@@ -136,17 +136,6 @@ static int tl_create(TabList *l)
     return 0;
 }
 
-static void ctu_tl_init(TabList *l, VVCFrameContext *fc)
-{
-    const VVCPPS *pps   = fc->ps.pps;
-    const int ctu_count = pps ? pps->ctb_count : 0;
-    const int changed   = fc->tab.sz.ctu_count != ctu_count;
-
-    tl_init(l, 1, changed);
-
-    TL_ADD(ctus,    ctu_count);
-}
-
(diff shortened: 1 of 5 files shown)


libavcodec/vvc/vvcdec.c
```diff
@@ -240,6 +275,19 @@ static void bs_tl_init(TabList *l, VVCFrameContext *fc)
...
+}
+
+static void bs_nz_tl_init(TabList *l, VVCFrameContext *fc)
+{
+    const VVCPPS *pps   = fc->ps.pps;
+    const int bs_width  = pps ? (pps->width >>  2) + 1 : 0;
+    const int bs_height = pps ? (pps->height >> 2) + 1 : 0;
+    const int bs_count  = bs_width * bs_height;
+    const int changed   = fc->tab.sz.bs_width != bs_width ||
+        fc->tab.sz.bs_height != bs_height;
```
QSXW, line 287: Is it better to align it with fc->tab.sz.bs_width != bs_width ||?
Like
```C++
const int changed   = fc->tab.sz.bs_width != bs_width ||
                      fc->tab.sz.bs_height != bs_height;
```
or not a new line. There are also long lines in the codes above.
```C++
const int changed   = fc->tab.sz.bs_width != bs_width || fc->tab.sz.bs_height != bs_height;
```

nuomi2021: sure, will do

nuomi2021: I combined bs with tu. so this line is not needed

PR #185, opened 2024-01-18
lavc/vvc: Fix left shift of negative value

frankplow:
`y_ctb` can be zero while `border_pixels`, `w` and `x` are all non-negative, resulting in a left shift of a negative value which is UB.  In the case `y_ctb` is 0, `edges[TOP]` is 1, hence the result is not used, but changing the order of operations removes the UB and makes undefined behaviour sanitizers happy.


nuomi2021:
thank you Frank. please send to upstream too.

PR #186, opened 2024-01-18
lavc/vvc: Fix MvField alignment

frankplow:
`IS_SAME_MV` requires motion vectors to be aligned to a multiple of 8 bytes:
https://github.com/ffvvc/FFmpeg/blob/ae5efc8e4b4d3ea0ce673d0ee69e425487591d42/libavcodec/vvc/vvc_mvs.c#L28
This PR adds the same requirement to the definition of MvField.  It fixes errors with UB sanitizers.


nuomi2021:
thank you please send it to upstream too. 
If you have a sanitizer error link, maybe you can attach it to commit log

PR #191, opened 2024-02-01
lavc/vvc: Support rectangular single-slice subpics

frankplow:
This PR adds support for rectangular single-slice subpictures, as used by SUBPIC_C_ERICSSON_1 and SUBPIC_D_ERICSSON_1.  Before these changes, these bitstreams would crash with an out-of-bounds array access.  At the time of writing, the files now decode, but do not produce the expected output.  Intra frames appear to decode correctly, however inter frames do not.  Visually, the nature of these distortions are similar to those in LMCS_B_Dolby_2 — @nuomi2021 do you think perhaps there is some bug with inter prediction that we have not uncovered yet?  It seems very unlikely that the bug would only affect these bitstreams which were failing for other reasons though.

### DIFF
--- a/libavcodec/cbs_h266_syntax_template.c
+++ b/libavcodec/cbs_h266_syntax_template.c
@@ -2118,6 +2118,8 @@ static int FUNC(pps) (CodedBitstreamContext *ctx, RWContext *rw,
             if (current->pps_no_pic_partition_flag)
                 infer(pps_num_slices_in_pic_minus1, 0);
             else if (current->pps_single_slice_per_subpic_flag)
+                for (i = 0; i <= sps->sps_num_subpics_minus1; i++)
+                    current->num_slices_in_subpic[i] = 1;
                 infer(pps_num_slices_in_pic_minus1,
                       sps->sps_num_subpics_minus1);
             // else?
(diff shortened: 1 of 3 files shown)


nuomi2021:
Ok, you are so fast, I think it's a good GOC project. and invite you comentor it :)
Will check it this weekend.

frankplow:
> Ok, you are so fast, I think it's a good GOC project. and invite you comentor it :) Will check it this weekend.

Lol sorry, just saw that.  I was working on this last night because I saw SUBPIC_C_ERICSSON_1 and SUBPIC_D_ERICSSON_1 were crashes rather than just mismatches.  I also have an alternative patch that throws an AVERROR_PATCHWELCOME if the code path that results in the out-of-bounds access will be activated if you'd rather merge that and leave the implementation for a GSoC student.

nuomi2021:
> do you think perhaps there is some bug with inter prediction that we have not uncovered yet?

Need to check the spec. But maybe because the subpicture will not cross subpicture boundary. 
It's not so easy, let use do it in GSOC :)

nuomi2021:
Since missed subpic code will crash the ffmpeg. So we can't wait for GSOC. 
I have sent the implementation code to upstream.  Please use this patchset
https://patchwork.ffmpeg.org/project/ffmpeg/patch/TYSPR06MB6433035C09F62A3DBF36D2B5AA2D2@TYSPR06MB6433.apcprd06.prod.outlook.com/

thank you

PR #216, opened 2024-04-20
Fix init task

nuomi2021:


nuomi2021:
@frankplow , this will fix all three files you send me. please help review.

frankplow:
@nuomi2021 LGTM!  Can confirm it fixes the hang with those bitstreams, plus also another ~dozen bitstreams I had which were causing hangs.  Don't have any bitstreams which cause hangs now.

frankplow:
@nuomi2021 Actually it looks like while this patch fixes most hangs, it causes a regression which makes some other bitstreams hang.  For example [id:000132](https://streams.videolan.org/ffmpeg/incoming/id%3A000132%2Csig%3A06%2Csrc%3A005246%2Ctime%3A670094688%2Cexecs%3A293703%2Cop%3Ahavoc%2Crep%3A3) fails on the assertion
https://github.com/ffvvc/FFmpeg/blob/e81b6d78fc2ddf8edd53a6a052713354ef8d27c2/libavcodec/vvc/vvc_thread.c#L182
before the patch, but hangs after the patch.

nuomi2021:
hmm, could you help https://github.com/ffvvc/FFmpeg/issues/218 and add all fixed fuzz clips to failed/fuzz/ ?
thank you

nuomi2021:
@frankplow how about the latest one :)

frankplow:
@nuomi2021 Yes, seems to have done the trick!

nuomi2021:
@frankplow clip name is not valid on Windows :)

frankplow:
> @frankplow clip name is not valid on Windows :)

Fixed in https://github.com/ffvvc/tests/pull/32

PR #233, opened 2024-06-02
draft (wip) - vvc_deblocking - chroma

stone-d-chen:
Started taking some notes on the differences between the hevc and vvc implementation.

The current plan is to implement the loop_filter_luma_large as a separate asm chunk re-using bits of code for the hevc_deblock.asm

The hevc_deblock.asm hooks hevc_h_loop_filter_chroma which integrates the decision for filter selection and the calculation. This makes it a bit complicated since vvc has new decisions to make (large filter vs original small), the original small filter has slightly different inputs between hevc / vvc, hevc needs to be ported to avx2, and generally my lack of knowledge of deblocking.

Instead the idea is to start with just the loop_filter_luma_large and hook it directly first. I can re-use utility functions from the HEVC implementation (while porting to avx2) and it will be more easily testable. Once that's completed, I should have enough knowledge to work on integrating the original hevc code.

### DIFF
--- a/tests/checkasm/vvc_deblock.c
+++ b/tests/checkasm/vvc_deblock.c
@@ -18,48 +18,29 @@
 
 #include <string.h>
 
-#include "checkasm.h"
-#include "libavcodec/vvc/ctu.h"
-#include "libavcodec/vvc/data.h"
 #include "libavcodec/vvc/dsp.h"
 
 #include "libavutil/common.h"
 #include "libavutil/intreadwrite.h"
 #include "libavutil/mem_internal.h"
 
+#include "checkasm.h"
+
 static const uint32_t pixel_mask[3] = {0xffffffff, 0x03ff03ff, 0x0fff0fff};
 
 #define SIZEOF_PIXEL ((bit_depth + 7) / 8)
 #define BUF_SIZE 16 * 16
 
-#define randomize_buffers(buf0, buf1, size)               \
-    do                                                    \
-    {                                                     \
-        uint32_t mask = pixel_mask[(bit_depth - 8) >> 1]; \
-        int k;                                            \
-        for (k = 0; k < size; k += 4 / sizeof(*buf0))     \
-        {                                                 \
-            uint32_t r = rnd() & mask;                    \
-            AV_WN32A(buf0 + k, r);                        \
-            AV_WN32A(buf1 + k, r);                        \
-        }                                                 \
+#define randomize_buffers(buf0, buf1, size)                 \
+    do {                                                    \
+        uint32_t mask = pixel_mask[(bit_depth - 8) >> 1];   \
+        int k;                                              \
+        for (k = 0; k < size; k += 4) {                     \
+            uint32_t r = rnd() & mask;                      \
+            AV_WN32A(buf0 + k, r);                          \
+            AV_WN32A(buf1 + k, r);                          \
+        }                                                   \
     } while (0)
-
-#define TC25(x) ((tc[x] * 5 + 1) >> 1)
-#define MASK(x) (uint16_t)(x & ((1 << (bit_depth)) - 1))
-#define GET(x) ((SIZEOF_PIXEL == 1) ? *(uint8_t *)(&x) : *(uint16_t *)(&x))
-#define SET(x, y)                  \
-    do                             \
-    {                              \
-        uint16_t z = MASK(y);      \
-        if (SIZEOF_PIXEL == 1)     \
-            *(uint8_t *)(&x) = z;  \
-        else                       \
-            *(uint16_t *)(&x) = z; \
-    } while (0)
-#define RANDCLIP(x, diff) av_clip(GET(x) - (diff), 0,       \
-                                  (1 << (bit_depth)) - 1) + \
-                              rnd() % FFMAX(2 * (diff), 1)
 
 #define P3 buf[-4 * xstride]
 #define P2 buf[-3 * xstride]
@@ -70,8 +51,64 @@ static const uint32_t pixel_mask[3] = {0xffffffff, 0x03ff03ff, 0x0fff0fff};
 #define Q2 buf[2 * xstride]
 #define Q3 buf[3 * xstride]
 
-static void check_deblock_chroma_horizontal()
-{   
+#define TC25(x) ((tc[x] * 5 + 1) >> 1)
+#define MASK(x) (uint16_t)(x & ((1 << (bit_depth)) - 1))
+#define GET(x) ((SIZEOF_PIXEL == 1) ? *(uint8_t*)(&x) : *(uint16_t*)(&x))
+#define SET(x, y) do { \
+    uint16_t z = MASK(y); \
+    if (SIZEOF_PIXEL == 1) \
+        *(uint8_t*)(&x) = z; \
+    else \
+        *(uint16_t*)(&x) = z; \
+} while (0)
+#define RANDCLIP(x, diff) av_clip(GET(x) - (diff), 0, \
+    (1 << (bit_depth)) - 1) + rnd() % FFMAX(2 * (diff), 1)
+static void randomize_chroma_buffers(int type, int beta[4], int32_t tc[4],
+   uint8_t *buf, ptrdiff_t xstride, ptrdiff_t ystride, const int shift, const int bit_depth)
+{
+    int i, j, b3, tc25, tc25diff, b3diff;
+    const int size = shift ? 4 : 2;
+    const int end = 8 / size;
+
+    for (j = 0; j < size; j++)
+    {
+        tc25 = TC25(j) << (bit_depth - 10);
+
+        tc25diff = FFMAX(tc25 - 1, 0);
+        // 2 or 4 lines per tc
+        for (i = 0; i < end; i++)
+        {
+            b3 = (beta[j] << (bit_depth - 8)) >> 3;
+
+            SET(P0, rnd() % (1 << bit_depth));
+            SET(Q0, RANDCLIP(P0, tc25diff));
+
+            // p3 - p0 up to beta3 budget
+            b3diff = rnd() % b3;
+            SET(P3, RANDCLIP(P0, b3diff));
+            // q3 - q0, reduced budget
+            b3diff = rnd() % FFMAX(b3 - b3diff, 1);
+            SET(Q3, RANDCLIP(Q0, b3diff));
+
+            // same concept, budget across 4 pixels
+            b3 -= b3diff = rnd() % FFMAX(b3, 1);
+            SET(P2, RANDCLIP(P0, b3diff));
+            b3 -= b3diff = rnd() % FFMAX(b3, 1);
+            SET(Q2, RANDCLIP(Q0, b3diff));
+
+            // extra reduced budget for weighted pixels
+            b3 -= b3diff = rnd() % FFMAX(b3 - (1 << (bit_depth - 8)), 1);
+            SET(P1, RANDCLIP(P0, b3diff));
+            b3 -= b3diff = rnd() % FFMAX(b3 - (1 << (bit_depth - 8)), 1);
+            SET(Q1, RANDCLIP(Q0, b3diff));
+
+            buf += ystride;
+        }
+    }
+}
+
+static void check_deblock_chroma(const VVCDSPContext *h, int bit_depth)
+{
     int32_t tc[4] = {(rnd() & 393) + 3, 600, (rnd() & 393) + 3, (rnd() & 393) + 3};
 
     uint8_t no_p[4] = {0, 0, 0, 0};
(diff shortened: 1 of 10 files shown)


stone-d-chen:
Wow just kidding, I missed `ff_vvc_deblock_horizontal` and `ff_vvc_deblock_vertical`

nuomi2021:
change hevc c code is risky,
How about we do things like this
1. copy or reuse hevc's check asm
2. make the current hevc test case passed for vvc
3. add new cases for loop_filter_chroma_strong
4. implement loop_filter_chroma_strong AVX2
5. continue steps 3 to 4 for  loop_filter_chroma_strong_one_side, loop_filter_luma_large

stone-d-chen:
> change hevc c code is risky, How about we do things like this

Oh yeah, I was just moving it around so I could visually compare the differences between the vvc & hevc implementations more easily

> 1. copy or reuse hevc's check asm
> 2. make the current hevc test case passed for vvc
> 3. add new cases for loop_filter_chroma_strong
> 4. implement loop_filter_chroma_strong AVX2
> 5. continue steps 3 to 4 for  loop_filter_chroma_strong_one_side, loop_filter_luma_large

Sounds good! I'll start this

nuomi2021:
> 
> Oh yeah, I was just moving it around so I could visually compare the differences between the vvc & hevc implementations more easily
> 

you can use https://winmerge.org/ or other tools to compare

stone-d-chen:
> > Oh yeah, I was just moving it around so I could visually compare the differences between the vvc & hevc implementations more easily
> 
> you can use https://winmerge.org/ or other tools to compare

Yeah I ended buying a copy of Araxis merge since that's what I use at work haha.

Update: I hacked in the chroma_weak asm and it seems to be working on conformance - just via c&p. I'm going to switch it to AVX2, debug, and then finish adding in the checkasm.

nuomi2021:
could we change the hevc deblock ASM code directly? 
we can add code like this, and define VVC to 1 when we use it for vvc
```
%if VVC
   loop_filter_chroma_strong 
%endif

```

stone-d-chen:
Yeah I think that works, the end goal is to share a single file for hevc and vvc?

nuomi2021:
Yes. We can renmae the hevc_deblock to h2645_deblock.asm.  most code are there.
in the vvc_deblock.asm, we only need two lines
```
%include h2645_deblock.asm
LOOP_FILTER vvc
```

in the hevc_deblock.asm, we also only need two lines
```
%include h2645_deblock.asm
LOOP_FILTER hevc
```

nuomi2021:
If possible, check_asm can share one function too. You can think about it :)

nuomi2021:
> Yes. We can renmae the hevc_deblock to h2645_deblock.asm. most code are there. in the vvc_deblock.asm, we only need two lines
> 
> ```
> %include h2645_deblock.asm
> LOOP_FILTER vvc
> ```
> 
> in the hevc_deblock.asm, we also only need two lines
> 
> ```
> %include h2645_deblock.asm
> LOOP_FILTER hevc
> ```

Feel free to schedule a call, if anything is not clear to you

stone-d-chen:
> Yes. We can renmae the hevc_deblock to h2645_deblock.asm. most code are there. in the vvc_deblock.asm, we only need two lines
> 
> ```
> %include h2645_deblock.asm
> LOOP_FILTER vvc
> ```
> 
> in the hevc_deblock.asm, we also only need two lines
> 
> ```
> %include h2645_deblock.asm
> LOOP_FILTER hevc
> ```

Yeah that makes sense - I added in the final pieces to adapt chroma_weak for 8 and 12 bits in the current files.

The next piece I am working on is translating the following to asm https://github.com/ffvvc/FFmpeg/blob/e81b6d78fc2ddf8edd53a6a052713354ef8d27c2/libavcodec/vvc/vvc_filter_template.c#L713-L754

So then I can add in the test cases for the strong and one-side strong into the same code path. 

I was thinking another approach would be to just start working on the strong/one-sided strong, and work on integrating the cases later.

stone-d-chen:
Also I was looking at how the HEVC luma deblocking works on calculating the weak / strong filter and uses masks to the masks - I think I'll need to do something similar for the chroma strong / one-sided strong.

> Feel free to schedule a call, if anything is not clear to you

Yeah it would be good to go over the general approach; I'm free most days 19:00 UTC+8 if that would work.

stone-d-chen:
Rough implementation of horizontal is done now - it has all the pieces necessary to handle the shift parameter as well

Rough benchmark
```
AVX:
 - vvc_deblock.chroma [OK]
checkasm: all 1 tests passed
vvc_h_loop_filter_chroma_10_c: 130.0
vvc_h_loop_filter_chroma_10_avx: 40.0
```

Doesn't seem to be an avx2 benefit which makes sense, not really exploiting the full register.

Next step is to clean-up the implementation then move to vertical / strong strong -- have a few days off from work so hoping to get this wrapped up by next week.

nuomi2021:
> vvc_h_loop_filter_chroma_10_c: 130.0
> vvc_h_loop_filter_chroma_10_avx: 40.0

👍, did you check the hevc performance, is it matched?

thank you

stone-d-chen:
> > vvc_h_loop_filter_chroma_10_c: 130.0
> > vvc_h_loop_filter_chroma_10_avx: 40.0
> 
> 👍, did you check the hevc performance, is it matched?
> 
> thank you

Hi, pulled some of the numbers below -- compared to hevc luma, it's very similar ~3-5x faster compared to the c version -- this is the appropriate comparison since we need to do weak/strong filtering in VVC chroma similar to hevc luma.


```
hevc_h_loop_filter_luma8_skip_c: 20.0
hevc_h_loop_filter_luma8_skip_avx: 9.7

hevc_h_loop_filter_luma8_strong_c: 160.0
hevc_h_loop_filter_luma8_strong_avx: 30.0

hevc_h_loop_filter_luma8_weak_c: 20.0
hevc_h_loop_filter_luma8_weak_avx: 10.0

hevc_h_loop_filter_luma10_strong_c: 160.0
hevc_h_loop_filter_luma10_strong_avx: 39.7

hevc_h_loop_filter_luma10_weak_c: 80.0
hevc_h_loop_filter_luma10_weak_avx: 30.0

hevc_h_loop_filter_luma12_skip_c: 29.7
hevc_h_loop_filter_luma12_skip_avx: 10.0

hevc_v_loop_filter_luma12_strong_c: 150.0
hevc_v_loop_filter_luma12_strong_avx: 39.7
```

stone-d-chen:
Quick update - horizontal is essentially done, I'm currently debugging an edge case failure, I think it's due to how I'm trying to test strong vs weak. I'm porting over the hevc checkasm code to generate more realistic test values. Once that's done, the vertical should be straightforward.

nuomi2021:
since the virtual boundaries patch merged. 
I have more time to check the code. Hope I can review the current patch before next week. 
thank you

stone-d-chen:
Sounds good, I fixed the issue (simple mistake forgetting to check max_len_q properly), I'll re-base on main soon. Next step is to clean up the checkasm (I did a messy hack to get the hevc code working).

stone-d-chen:
Update:

- checkasm has been updated with a specific strong deblock data generation -- 8 bit is a little broken, doesn't generate strong as often need to debug that
- Conformance doesn't pass actually, need to figure out what the issue is..
- I've separated out the Spatial Activity code into a separate macro for implementing the strong for vertical

stone-d-chen:
```
AVX:
 - vvc_deblock.chroma [OK]
AVX2:
 - vvc_deblock.chroma [OK]
checkasm: all 3 tests passed
vvc_h_loop_filter_chroma_8_c: 140.0
vvc_h_loop_filter_chroma_8_avx2: 40.0
vvc_h_loop_filter_chroma_10_c: 110.0
vvc_h_loop_filter_chroma_10_avx: 30.0
vvc_h_loop_filter_chroma_12_c: 120.0
vvc_h_loop_filter_chroma_12_avx: 40.0
```

tests/checkasm/vvc_deblock.c
```diff
@@ -0,0 +1,160 @@
...
+#define P1 buf[-2 * xstride]
+#define P0 buf[-1 * xstride]
+#define Q0 buf[0 * xstride]
+#define Q1 buf[1 * xstride]
+#define Q2 buf[2 * xstride]
+#define Q3 buf[3 * xstride]
+
+static void check_deblock_chroma_horizontal()
+{   
+    int32_t tc[4] = {(rnd() & 393) + 3, 600, (rnd() & 393) + 3, (rnd() & 393) + 3};
```
nuomi2021: why offset 3? https://github.com/ffvvc/FFmpeg/blob/e81b6d78fc2ddf8edd53a6a052713354ef8d27c2/libavcodec/vvc/vvc_filter.c#L38 has zeros

tests/checkasm/vvc_deblock.c
```diff
@@ -0,0 +1,160 @@
...
+
+    for (bit_depth = 8; bit_depth <= 12; bit_depth += 2)
+    {
+        int xstride = SIZEOF_PIXEL * 8; // bytes
+        int ystride = 2;                // bytes
+
+        ff_vvc_dsp_init(&context, bit_depth);
+        if (check_func(context.lf.filter_chroma[0], "vvc_h_loop_filter_chroma_%d", bit_depth))
+        {
+            int i, j, b3, tc25, tc25diff, b3diff;
```
nuomi2021: better move this to randomize_chroma_buffers. and follow the structure in https://github.com/ffvvc/FFmpeg/blob/e81b6d78fc2ddf8edd53a6a052713354ef8d27c2/tests/checkasm/hevc_deblock.c#L48

tests/checkasm/vvc_deblock.c
```diff
@@ -0,0 +1,160 @@
...
+            bench_new(buf0 + xstride * 5, xstride, beta, tc, no_p, no_q, max_len_p, max_len_q, shift);
+        }
+    }
+    report("chroma");
+}
+
+void checkasm_check_vvc_deblock(void)
+{
+    check_deblock_chroma_horizontal();
+}
```
nuomi2021: need a new line

libavcodec/vvc/filter.c
```diff
@@ -935,11 +942,21 @@ void ff_vvc_deblock_horizontal(const VVCLocalContext *lc, const int x0, const in
...
                     src = &fc->frame->data[c_idx][(y >> vs) * fc->frame->linesize[c_idx] + ((x >> hs) << fc->ps.sps->pixel_shift)];
                     if (!c_idx) {
-                        fc->vvcdsp.lf.filter_luma[0](src, fc->frame->linesize[c_idx],
-                            beta, tc, no_p, no_q, max_len_p, max_len_q, horizontal_ctu_edge);
+                        fc->vvcdsp.lf.filter_luma[0](src, fc->frame->linesize[c_idx], beta, tc, no_p, no_q, max_len_p, max_len_q, horizontal_ctu_edge);
                     } else {
-                        fc->vvcdsp.lf.filter_chroma[0](src, fc->frame->linesize[c_idx],
+                           
+                        if(
+                            max_len_p[0] == 1 && max_len_q[0] == 1 &&
```
nuomi2021: maybe we can set max_len_p and max_len_q to 1 in the check asm.

stone-d-chen: Yes, though this will eventually all go into the assembly to handle all cases.

stone-d-chen:
Update:
* Fixed a number of bugs that were causing conformance clips to not pass - focused on the shift version. The main one was the comparison could be off by 1 - roughly the issue were related to >= instead of strictly > ; so a somewhat difficult to catch bug with random values generated in checkasm.
* I didn't realize that Horizontal also used Strong -- it just additionally has one-sided strong; so it's not quite done yet.

Next steps:
* Fix conformance clip issues with non-shift version
* Carefully audit the comparison steps / generate edge case tests

nuomi2021:
No worries. 
Let's work together.I might send some PRs to your branch in the following days.

stone-d-chen:
Sounds good, one-sided strong/weak/spatial activity calculation are fully working now.

Going to start Strong and work out how to set up the masking properly. It's possible to have combinations of one-sided strong / strong / weak for horizontal (plus no deblocking).

stone-d-chen:
@nuomi2021 @QSXW 

One issue I'm running into, I actually am running out of registers to use as masks:

- Spatial activity generates 1, which determines the weak values (overriding strong) and then I need to determine Strong vs One stride strong (3 total masks)

I was thinking about using movmskps to save out the Spatial activity mask? Not sure if there's a better approach.

edit: Suggestion from IRC is to just spill to memory

nuomi2021:
Which line? yes, you an push to stack and pop it later.

stone-d-chen:
https://github.com/ffvvc/FFmpeg/blob/b63408057dd3b7374637665e8c3381ba8fae73d8/libavcodec/x86/vvc/vvc_deblock.asm#L728-L749

Roughly the SPATIAL_ACTIVITY macro creates a Weak vs (Strong/One-Sided) in m11 which I'm now saving to the stack.

This means I can continue using m11 for other masks ( the hevc MASKED_COPY macro uses m11).

I'll also save the no_p, no_q masks to the stack. That way we can see if any work needs to be done earlier and exit if not.

nuomi2021:
Ok, let me check it this weekend. 
could you help rebase your code against https://github.com/ffvvc/FFmpeg/tree/up
there are too many delta patches. pretty hard to review and comment
thank you

stone-d-chen:
> Ok, let me check it this weekend. could you help rebase your code against https://github.com/ffvvc/FFmpeg/tree/up there are too many delta patches. pretty hard to review and comment thank you

Rebased

Horizontal is mostly done finally! for shifted it handles any mix of strong/one-sided strong/weak 

- test non-shifted, there shouldn't be to much to do for this
- handle with max_len_p/q = 0
- move changes over to 8/12 bit

Once that's done vertical should be relatively straightforward since it will just need the SPATIAL_ACTIVITY macro and the strong/weak calcs.

There's still a lot of clean-up that can be done, I focused on getting things working so there's a lot of redundant ops.

```
skipped files:
    FIELD_B_Panasonic_2.bit
    FIELD_A_Panasonic_4.bit

total = 335, passed = 333, skipped = 2, failed = 0
```

stone-d-chen:
Horizontal is now correct for strong/one-sided strong/weak - the only case left is when max_len_p/q = 0

Performance has degraded quite a bit after adding strong - mainly because I'm doing all 3 computations every time and masking the results; next is checking the masks for all 0's to skip entire computations e.g. if there's no strong, skip all strong comps.

I'm also using a lot of shuffles to create the masks so there's probably some optimizations there.

```
checkasm: all 3 tests passed
vvc_h_loop_filter_chroma_8_c: 89.7
vvc_h_loop_filter_chroma_8_avx: 100.0
vvc_h_loop_filter_chroma_10_c: 90.0
vvc_h_loop_filter_chroma_10_avx: 90.0
vvc_h_loop_filter_chroma_12_c: 90.0
vvc_h_loop_filter_chroma_12_avx: 80.0
```

nuomi2021:
> There's still a lot of clean-up that can be done, I focused on getting things working so there's a lot of redundant ops.

no worries, I can send some clean-up pr to you. It will polish the code and help me understand more about your code.
thank you

nuomi2021:
> * handle with max_len_p/q = 0

if you think max_len_p and max_len_q = 0 are similar to no_q and np_p = 0. maybe we can change c code to remove no_q and no_p
> * move changes over to 8/12 bit

better use a unfied code for 8 to 12. or at least 10 to 12

nuomi2021:
@stone-d-chen , please close this and send a new pr to https://github.com/ffvvc/FFmpeg/tree/deblock_asm. 
It will help us focus on the detail commits.

PR #238, opened 2024-06-27
Use -Werror and -Wpedantic in CI

frankplow:
This patchset does two things:
* Adds the `-Werror` flag for only the libavcodec/vvc/*.o objects.  This turns any compilation warnings for these files into CI failures.
* Adds the `-Wpedantic` flag, which adds warnings whenever non-standard C features are used.  This should help catch errors such as that fixed by https://github.com/FFmpeg/FFmpeg/commit/75e3b81f75b778286db7257a0ec4af74913ea67f

Note these stricter compilation settings fail with the current HEAD, due to the error:
```
libavcodec/vvc/intra_template.c:201:63: error: pointers to arrays with different qualifiers are incompatible in ISO C [-Werror=pedantic]
  201 |     FUNC(cclm_select_luma)(fc, x0, y0, avail_t, avail_l, cnt, pos, sel[LUMA]);
```
in this location and various others.

### DIFF
--- a/.github/workflows/patches/20-vvc-werror.patch
+++ b/.github/workflows/patches/20-vvc-werror.patch
@@ -1,0 +1,23 @@
+diff --git a/libavcodec/vvc/Makefile b/libavcodec/vvc/Makefile
+index 6a28d32bc2..dc631897d4 100644
+--- a/libavcodec/vvc/Makefile
++++ b/libavcodec/vvc/Makefile
+@@ -15,3 +15,18 @@ OBJS-$(CONFIG_VVC_DECODER)          +=  vvc/dec.o           \
+                                         vvc/ps.o            \
+                                         vvc/refs.o          \
+                                         vvc/thread.o        \
++
++$(SUBDIR)vvc/dec.o: CFLAGS += -Werror
++$(SUBDIR)vvc/dsp.o: CFLAGS += -Werror
++$(SUBDIR)vvc/cabac.o: CFLAGS += -Werror
++$(SUBDIR)vvc/ctu.o: CFLAGS += -Werror
++$(SUBDIR)vvc/data.o: CFLAGS += -Werror
++$(SUBDIR)vvc/filter.o: CFLAGS += -Werror
++$(SUBDIR)vvc/inter.o: CFLAGS += -Werror
++$(SUBDIR)vvc/intra.o: CFLAGS += -Werror
++$(SUBDIR)vvc/intra_utils.o: CFLAGS += -Werror
++$(SUBDIR)vvc/itx_1d.o: CFLAGS += -Werror
++$(SUBDIR)vvc/mvs.o: CFLAGS += -Werror
++$(SUBDIR)vvc/ps.o: CFLAGS += -Werror
++$(SUBDIR)vvc/refs.o: CFLAGS += -Werror
++$(SUBDIR)vvc/thread.o: CFLAGS += -Werror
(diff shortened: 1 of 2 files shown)


frankplow:
> Note these stricter compilation settings fail with the current HEAD, due to the error:
> 
> ```
> libavcodec/vvc/intra_template.c:201:63: error: pointers to arrays with different qualifiers are incompatible in ISO C [-Werror=pedantic]
>   201 |     FUNC(cclm_select_luma)(fc, x0, y0, avail_t, avail_l, cnt, pos, sel[LUMA]);
> ```
> 
> in this location and various others.

This seems to be an issue with C more than anything else.  The issue is discussed in [N 1923](https://open-std.org/JTC1/SC22/wg14/www/docs/n1923.htm) in detail, including workarounds, and a fix has been accepted in C23 as [N 2607](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2607.pdf).  The workarounds don't sound ideal, so perhaps we explicitly ignore this warning?

.github/workflows/patches/20-vvc-werror.patch
```diff
@@ -0,0 +1,23 @@
+diff --git a/libavcodec/vvc/Makefile b/libavcodec/vvc/Makefile
+index 6a28d32bc2..dc631897d4 100644
+--- a/libavcodec/vvc/Makefile
++++ b/libavcodec/vvc/Makefile
+@@ -15,3 +15,18 @@ OBJS-$(CONFIG_VVC_DECODER)          +=  vvc/dec.o           \
+                                         vvc/ps.o            \
+                                         vvc/refs.o          \
+                                         vvc/thread.o        \
++
++$(SUBDIR)vvc/dec.o: CFLAGS += -Werror
```
nuomi2021: We may split or rename files. 
how about

`$(SUBDIR)vvc/*.o: CFLAGS += -Werror`

nuomi2021:
> This seems to be an issue with C more than anything else. The issue is discussed in [N 1923](https://open-std.org/JTC1/SC22/wg14/www/docs/n1923.htm) in detail, including workarounds, and a fix has been accepted in C23 as [N 2607](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2607.pdf). The workarounds don't sound ideal, so perhaps we explicitly ignore this warning?

Agree. It's an ISO C issue. :) We can ignore it.

frankplow:
> > This seems to be an issue with C more than anything else. The issue is discussed in [N 1923](https://open-std.org/JTC1/SC22/wg14/www/docs/n1923.htm) in detail, including workarounds, and a fix has been accepted in C23 as [N 2607](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2607.pdf). The workarounds don't sound ideal, so perhaps we explicitly ignore this warning?
> 
> Agree. It's an ISO C issue. :) We can ignore it.

Unfortunately there does not appear to be any easy way to do this.  There are no command line flags for GCC to selectively disable some ISO compliance checks, either it tests if the code is ISO C compliant or not.  The only workaround would be to use pragmas to selectively disable `-Wpedantic` for only the problematic lines, but we'd have to patch this in and the diff would break with nearly any change to the code so I don't think this is workable.  I think if we want to add these checks, our best bet is to modify the code to be fully ISO C compliant in these cases.

.github/workflows/patches/20-vvc-werror.patch
frankplow: Done in 723a9f02e25b5891d09cf6de78812b32c2347182.

nuomi2021:
>I think if we want to add these checks, our best bet is to modify the code to be fully ISO C compliant in these cases.

If this costs too much time, it may not be worth it.  Focus on fuzz may help us more. 

But if you want, we can change https://github.com/ffvvc/FFmpeg/blob/e81b6d78fc2ddf8edd53a6a052713354ef8d27c2/libavcodec/vvc/vvc_intra_template.c#L241 to 1d array like pixel sel[VVC_MAX_SAMPLE_ARRAYS * MAX_PICK_POS * 2];

Thank you

PR #239, opened 2024-07-07
vvc_h_loop_filter_chroma_{10, 12}, reuse CHROMA_DEBLOCK_BODY

nuomi2021:
@stone-d-chen, please help review.
also please help add a similar thing fro 8 bits, you can take hevc luma code as a reference.

thank you


tests/checkasm/vvc_deblock.c
```diff
@@ -64,11 +66,12 @@ static const uint32_t pixel_mask[3] = {0xffffffff, 0x03ff03ff, 0x0fff0fff};
...
     (1 << (bit_depth)) - 1) + rnd() % FFMAX(2 * (diff), 1)
 
-static void randomize_params(int32_t beta[4], int32_t tc[4], const int size)
+static void randomize_params(int32_t beta[4], int32_t tc[4], const int bit_depth, const int size)
 {
+    const int tc_min = 1 << FFMAX(0, 9 - bit_depth);
     for (int i = 0; i < size; i++) {
         beta[i] = rand() % 89;
-        tc[i] = rand() % 396;
+        tc[i] = (rand() % (396 - tc_min)) + tc_min;   // minimum useful value is tc_min, full range [0, 395]
```
stone-d-chen: I think rand should be rnd(), I don't think we get different values for different checkasm runs otherwise.

Beta value can become zero which will throw an exception in a later line `b3diff = rnd() % b3;`

We can probably update that to `b3diff = rnd() % FFMAX(b3, 1);`

Alternatively we can update beta value as well

Similar to hevc_deblock.c
 // minimum useful value for 8bit is 8
    *beta = (rnd() % 57) + 8;

nuomi2021: Thank you for the review.
 I will send another PR for this.  
Why is weak not tested in the current vvc_deblock.c? Based on the code, I think starting from weak -> one_side_strong -> strong will make our work easier. because the one side and weak are decided by max_len_{p,q} only

stone-d-chen: > Why is weak not tested in the current vvc_deblock.c? 

Artifact of focusing on debugging strong code (also laziness ), I'll be sending an updated version soon that tests strong, one-side, weak, mix for shift / no shift.

> Based on the code, I think starting from weak -> one_side_strong -> strong will make our work easier. because the one side and weak are decided by max_len_{p,q} only

Only weak is decided by max_len_{p,q} only, one-sided still goes through the spatial activity calculation to determine if it gets reduced to weak (similar to strong).

https://github.com/ffvvc/FFmpeg/blob/8c8fd4e85dc7872baf31db4c2dce2f034cd11a31/libavcodec/vvc/filter_template.c#L707-L713

PR #244, opened 2024-07-13
checkasm: vvc_deblock, fix tc,beta generator for VVC

nuomi2021:
@stone-d-chen , 
if you check the luma and chroma code, 
the beta == 0 will disable luma filter. but beta == 0 will not disable chroma filter.
so beta == 0 is valid for chroma. maybe we need to fix the randomize_chroma_buffers

thank you

### DIFF
--- a/tests/checkasm/vvc_deblock.c
+++ b/tests/checkasm/vvc_deblock.c
@@ -23,6 +23,7 @@
 #include "libavutil/common.h"
 #include "libavutil/intreadwrite.h"
 #include "libavutil/mem_internal.h"
+#include "libavutil/macros.h"
 
 #include "checkasm.h"
 
@@ -51,7 +52,8 @@ static const uint32_t pixel_mask[3] = {0xffffffff, 0x03ff03ff, 0x0fff0fff};
 #define Q2 buf[2 * xstride]
 #define Q3 buf[3 * xstride]
 
-#define TC25(x) ((tc[x] * 5 + 1) >> 1)
+#define TC_SHIFT(j) ((bit_depth < 10) ? ((tc[j] + (1 << (9 - bit_depth))) >> (10 - bit_depth)) : (tc[j] << (bit_depth - 10)))
+#define TC25(x) ((TC_SHIFT(j) * 5 + 1) >> 1)
 #define MASK(x) (uint16_t)(x & ((1 << (bit_depth)) - 1))
 #define GET(x) ((SIZEOF_PIXEL == 1) ? *(uint8_t*)(&x) : *(uint16_t*)(&x))
 #define SET(x, y) do { \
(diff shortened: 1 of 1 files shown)


tests/checkasm/vvc_deblock.c
```diff
@@ -64,11 +66,13 @@ static const uint32_t pixel_mask[3] = {0xffffffff, 0x03ff03ff, 0x0fff0fff};
...
-static void randomize_params(int32_t beta[4], int32_t tc[4], const int size)
+static void randomize_params(int32_t beta[4], int32_t tc[4], const int is_luma, const int bit_depth, const int size)
 {
+    const int tc_min   = 1 << FFMAX(0, 9 - bit_depth);
+    const int beta_min = is_luma;                     // for luma, beta == 0 will disable the deblock, for chroma beta == 0 is a valid value
     for (int i = 0; i < size; i++) {
-        beta[i] = rand() % 89;
-        tc[i] = rand() % 396;
+        beta[i] = rand() % (89 - beta_min) + beta_min;
+        tc[i] = (rand() % (396 - tc_min)) + tc_min;   // minimum useful value is tc_min, full range [0, 395]
```
stone-d-chen: `rand()` should be `rnd()`

Would this be easier?
```
beta[i] = FFMAX(rnd() % 89, beta_min);
tc[i] = FFMAX(rnd() % 396, tc_min);
```

tests/checkasm/vvc_deblock.c
```diff
@@ -51,7 +52,8 @@ static const uint32_t pixel_mask[3] = {0xffffffff, 0x03ff03ff, 0x0fff0fff};
 #define Q2 buf[2 * xstride]
 #define Q3 buf[3 * xstride]
 
-#define TC25(x) ((tc[x] * 5 + 1) >> 1)
+#define TC_SHIFT(j) ((bit_depth < 10) ? ((tc[j] + (1 << (9 - bit_depth))) >> (10 - bit_depth)) : (tc[j] << (bit_depth - 10)))
+#define TC25(x) ((TC_SHIFT(j) * 5 + 1) >> 1)
```
stone-d-chen, line 56: This probably should be ported over to hevc_deblock.c as well?

tests/checkasm/vvc_deblock.c
nuomi2021: > rand() should be rnd()

sorry for missing this

> Would this be easier?

good idea, fixed

tests/checkasm/vvc_deblock.c
nuomi2021: if you check the code, hevc will use the same beta. but vvc did no.
you can use the same beta in vvc code. If all conformance passed, maybe we can use the same beta too

nuomi2021:
Let me merge this.
We can check the HEVC status later before we send the code upstream. We'll do our best to compare, refactor, and reuse code from HEVC.

PR #259, opened 2024-08-24
Vvc v loop filter chroma

nuomi2021:
@stone-d-chen ,

Thilo has requested that we submit the deblocking filter upstream for review. Could you help by cherry-picking this and sending out the chroma part this week?

The chroma BODY still needs some refactoring to align with HEVC and to reduce/remove stack operations like the current luma. I'll work with you on this, but it’s fine to send it out as the first version for now.

thank you

### DIFF
--- a/libavcodec/x86/vvc/vvc_deblock.asm
+++ b/libavcodec/x86/vvc/vvc_deblock.asm
@@ -122,26 +122,22 @@ SECTION .text
     movu             %8, m7
 %endmacro
 
-
 ; in: %2 clobbered
 ; out: %1
-; mask in m11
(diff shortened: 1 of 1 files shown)


stone-d-chen:
hi @nuomi2021 

Starting to look at ffmpeg stuff again, do you still want me to cherry pick and send it to the ML? Otherwise I can start looking at the refactoring.

nuomi2021:
Welcome back!
We still need to refactor the code to share more with HEVC. The chroma component can share 100% of the code with HEVC; we just need to use some macros to separate the VVC parts. You can use the luma component as an example to compare the difference
We need to move this shared to h26x/

PR #267, opened 2024-10-26
Share VVC/HEVC deblock code (weak chroma / chroma)

stone-d-chen:
Given that HEVC chroma deblock is just the weak subset of VVC chroma deblock, attempt to share code

* Attempt to replace WEAK_CHROMA (vvc) with CHROMA_DEBLOCK_BODY (vvc)

### DIFF
--- a/libavcodec/x86/h26x/h2656_deblock.asm
+++ b/libavcodec/x86/h26x/h2656_deblock.asm
@@ -670,3 +670,42 @@
     psubw            m8, m4, m12 ; q0 - delta0
     MASKED_COPY      m4, m8
 %endmacro
+
+%macro H2656_CHROMA_DEBLOCK 13
+    psubw            %5, %4, %3; q0 - p0
+    psubw            %8, %7, %6; p1 - q1
+    psllw            %5, 2; << 2
+    paddw            %8, %5;
+
+%ifidn %2, hevc
+    ;tc calculations
+    movq             m6, [tcq]; tc0
+    punpcklwd        m6, m6
+    pshufd           m6, m6, 0xA0; tc0, tc1
+%if cpuflag(ssse3)
+    psignw           m4, m6, [pw_m1]; -tc0, -tc1
+%else
+    pmullw           m4, m6, [pw_m1]; -tc0, -tc1
+%endif
+    ;end tc calculations
+%if %1 > 8
+    psllw            m4, %1-8; << (BIT_DEPTH - 8)
+    psllw            m6, %1-8; << (BIT_DEPTH - 8)
+%endif
+%endif
+
+    paddw            %8, [pw_4]; +4
+    psraw            %8, 3; >> 3
+
+%ifidn %2, hevc
+    pmaxsw           %8, %5
+    pminsw           %8, %9
+    paddw            %3, %8; p0 + delta0
+    psubw            %4, %8; q0 - delta0
+%elifidn %2, vvc
+    pmaxsw           %8, %12
+    pminsw           %8, %13
+    paddw           %10,  %3, %8
+    psubw           %11,  %4, %8
+%endif
+%endmacro
--- a/libavcodec/x86/hevc_deblock.asm
+++ b/libavcodec/x86/hevc_deblock.asm
@@ -46,34 +46,8 @@ INIT_XMM sse2
 
 ALIGN 16
 ; input in m0 ... m3 and tcs in r2. Output in m1 and m2
-%macro CHROMA_DEBLOCK_BODY 1
-    psubw            m4, m2, m1; q0 - p0
-    psubw            m5, m0, m3; p1 - q1
-    psllw            m4, 2; << 2
-    paddw            m5, m4;
-
-    ;tc calculations
-    movq             m6, [tcq]; tc0
-    punpcklwd        m6, m6
-    pshufd           m6, m6, 0xA0; tc0, tc1
-%if cpuflag(ssse3)
-    psignw           m4, m6, [pw_m1]; -tc0, -tc1
-%else
-    pmullw           m4, m6, [pw_m1]; -tc0, -tc1
-%endif
-    ;end tc calculations
-
-    paddw            m5, [pw_4]; +4
-    psraw            m5, 3; >> 3
-
-%if %1 > 8
-    psllw            m4, %1-8; << (BIT_DEPTH - 8)
-    psllw            m6, %1-8; << (BIT_DEPTH - 8)
-%endif
-    pmaxsw           m5, m4
-    pminsw           m5, m6
-    paddw            m1, m5; p0 + delta0
-    psubw            m2, m5; q0 - delta0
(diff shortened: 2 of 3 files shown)


stone-d-chen:
Hi @nuomi2021 

Started looking at sharing the VVC/HEVC code, I'm not sure how easy it is without changing the CHROMA_DEBLOCK_BODY from hevc. I could be misunderstanding what you were suggesting though.

Two issues I'm running into
* in vvc I need to do  `paddw            m14, m3, m13` since I can't clobber m3; It can store non-weak data.
https://github.com/ffvvc/FFmpeg/blob/79a17dd3f8bd547eed209f71f0a9c77e88e82b67/libavcodec/x86/vvc/vvc_deblock.asm#L180-L184

https://github.com/ffvvc/FFmpeg/blob/79a17dd3f8bd547eed209f71f0a9c77e88e82b67/libavcodec/x86/hevc_deblock.asm#L300-L303

* I used different registers (and a lot more) in VVC leading up to the weak filter; I tried solving this with SWAP but it seems quite tricky to get right. So the only easy solution would be to convert CHROMA_DEBLOCK_BODY into a macro, with a parameter for each register.

```
; pseudo-code
%macro CHROMA_DEBLOCK_BODY 7
    psubw            %5, %3, %2; q0 - p0
    psubw            %6, %1, %4; p1 - q1
    psllw            %6 2; << 2
    paddw            %6, %5;
```

I could probably get around this be re-writing a single macro to share between VVC/HEVC - this should be doable but I'm a little nervous about modifying the HEVC version, especially given how few lines of code are really shared. That is I'm not sure the risk of needing to modify the HEVC version to share code is worth it. 

Let me know what you think/if I'm misunderstanding something here.

Thanks!

nuomi2021:
Just like this, but starting from HEVC makes it easier.

Take [this pull request](https://github.com/ffvvc/FFmpeg/pull/268) as an example:

1. Move HEVC code to h26x, make chekasm and fate-hevc passed.
2. Make necessary changes for VVC
3. Update the VVC decoder to use the h26x code
4. Add checkasm code

Following these steps also helps with the upstream review process.

nuomi2021:
@stone-d-chen   I noticed that chroma is still pushing some memory to the stack, which luma doesn’t require. Do we really need them?

stone-d-chen:
> Just like this, but starting from HEVC makes it easier.
> 
> Take [this pull request](https://github.com/ffvvc/FFmpeg/pull/268) as an example:
> 
> 1. Move HEVC code to h26x, make chekasm and fate-hevc passed.

Oh awesome I didn't realize the fate-hevc existed.


> @stone-d-chen I noticed that chroma is still pushing some memory to the stack, which luma doesn’t require. Do we really need them?

Mm the "shift" and other things in vvc chroma causes some complications -- For Example: only for the shifted samples, the spatial activity calculation uses P1 instead of P3/P2. However the actual filter values still will use P3/P2 even if shifted. Since I need to calculate the entire blocks spatial activity at once for performance, I need an simd register where some are modified as necessary (MASKED COPY) and then restore the original sample at some point. For blocks not transposed, this is easy, I can just `movu m0 ...` but for transposed versions, I need to store the transposed samples somewhere. Other stuff like this causes similar stack use.

That is, I think it's possible to remove the stack use but it might be quite involved.

https://github.com/ffvvc/FFmpeg/blob/79a17dd3f8bd547eed209f71f0a9c77e88e82b67/libavcodec/x86/vvc/vvc_deblock.asm#L285-L293

nuomi2021:
> Mm the "shift" and other things in vvc chroma causes some complications -- For Example: only for the shifted samples, the spatial activity calculation uses P1 instead of P3/P2. However the actual filter values still will use P3/P2 even if shifted. Since I need to calculate the entire blocks spatial activity at once for performance, I need an simd register where some are modified as necessary (MASKED COPY) and then restore the original sample at some point. For blocks not transposed, this is easy, I can just `movu m0 ...` but for transposed versions, I need to store the transposed samples somewhere. Other stuff like this causes similar stack use.
> 
> That is, I think it's possible to remove the stack use but it might be quite involved.

Need time to read more of your code to give you some suggestion.

stone-d-chen:
Update: Started making some changes to hevc but struggling a bit to get fate for hevc to build properly (and windows), there's some patches on the ML I'm going to try and apply, otherwise will switch to Linux.

nuomi2021:
> Update: Started making some changes to hevc but struggling a bit to get fate for hevc to build properly (and windows), there's some patches on the ML I'm going to try and apply, otherwise will switch to Linux.

if you are on a windows, you can use WSL2.

stone-d-chen:
> > Update: Started making some changes to hevc but struggling a bit to get fate for hevc to build properly (and windows), there's some patches on the ML I'm going to try and apply, otherwise will switch to Linux.
> 
> 
> 
> if you are on a windows, you can use WSL2. 

fate-hevc and deblock checkasm passed with the changes.

stone-d-chen:
```
$ make checkasm && ./tests/checkasm/checkasm.exe --test=hevc_deblock && make fate-hevc       SAMPLES=fate-suite/
make: Nothing to be done for 'checkasm'.
checkasm: using random seed 3697830752
SSE2:
 - hevc_deblock.chroma      [OK]
 - hevc_deblock.luma        [OK]
SSSE3:
 - hevc_deblock.luma        [OK]
AVX:
 - hevc_deblock.chroma      [OK]
 - hevc_deblock.luma        [OK]
checkasm: all 66 tests passed
TEST    hevc-paired-fields
TEST    hevc-monochrome-crop
TEST    hevc-hdr10-plus-metadata
TEST    hevc-hdr-vivid-metadata
```

Jamaika1:
Hi,
Rhetorical questions
Does movie have errors or is it a decoder issue?
https://www.sendspace.com/file/tifjuj
AVX2 for hevc doesn't work for me in the latest ffmpeg 7.1.0. I don't know where the reason might be. Maybe decoder is being rebuilt. I added these changes, but that's not it.

stone-d-chen:
> Hi, Rhetorical questions Does movie have errors or is it a decoder issue? https://www.sendspace.com/file/tifjuj AVX2 for hevc doesn't work for me in the latest ffmpeg 7.1.0. I don't know where the reason might be. Maybe decoder is being rebuilt. I added these changes, but that's not it.

Hi @Jamaika1,

Any HEVC work in this repo isn't changing the current functionality of the HEVC decoder, just refactoring it. I'm not sure if there have been other changes. You can file a bug report https://trac.ffmpeg.org/

nuomi2021:
@stone-d-chen , could you help rebase the code to the latest https://github.com/ffvvc/FFmpeg/tree/deblock_asm. 

Thank you

stone-d-chen:
> @stone-d-chen , could you help rebase the code to the latest https://github.com/ffvvc/FFmpeg/tree/deblock_asm.
> 
> Thank you

Hi @nuomi2021, this has been rebased

nuomi2021:
Thank you. Stone.
Will spend time to read detail of your code and try to remove stack if it's possible.

PR #273, opened 2024-11-24
vvc_deblock.c: fix checkasm for one-sided chroma

stone-d-chen:
Missing an spatial_strong <= meant that the one-sided never ran properly.


stone-d-chen:
@nuomi2021 Should be good to go now, I forgot I had code that turned off the one-sided code path for vertical :sweat_smile:

PR #274, opened 2024-12-01
Simplfy deblock chroma strong

nuomi2021:


nuomi2021:
@stone-d-chen, there are still some issues with our checkasm test for luma.
If I add the following code, the checkasm test passes, but ALF_B_Huawei_3.bit and several other cases fail.
I will dig more, could you help check it from test case side?
Thank you

```
diff --git a/libavcodec/vvc/filter.c b/libavcodec/vvc/filter.c
index 069a111987..d3be8599f8 100644
--- a/libavcodec/vvc/filter.c
+++ b/libavcodec/vvc/filter.c
@@ -838,7 +838,7 @@ static void vvc_deblock(const VVCLocalContext *lc, int x0, int y0, const int rs,
                     if (!c_idx) {
                         int use_c = 0;
                         for (int i = 0; i < 2; i++) {
-                            if (max_len_p[i] > 3 || max_len_q[i])
+                            if (max_len_p[i] > 3 || max_len_q[i] > 1)
                                 use_c = 1;
                         }
                         if (use_c) {
```

stone-d-chen:
> @stone-d-chen, there are still some issues with our checkasm test for luma. If I add the following code, the checkasm test passes, but ALF_B_Huawei_3.bit and several other cases fail. I will dig more, could you help check it from test case side? Thank you
> 
> ```
> diff --git a/libavcodec/vvc/filter.c b/libavcodec/vvc/filter.c
> index 069a111987..d3be8599f8 100644
> --- a/libavcodec/vvc/filter.c
> +++ b/libavcodec/vvc/filter.c
> @@ -838,7 +838,7 @@ static void vvc_deblock(const VVCLocalContext *lc, int x0, int y0, const int rs,
>                      if (!c_idx) {
>                          int use_c = 0;
>                          for (int i = 0; i < 2; i++) {
> -                            if (max_len_p[i] > 3 || max_len_q[i])
> +                            if (max_len_p[i] > 3 || max_len_q[i] > 1)
>                                  use_c = 1;
>                          }
>                          if (use_c) {
> ```

Sorry for the delay, just had a chance to look at it,

Seems like the checkasm doesn't trigger the `loop_filter_luma_large` pathway

DMVR_B_KDDI_4.bit does not use the loop_filter_luma_large path and will pass when using asm.

PR #277, opened 2025-01-28
Simplify deblock chroma tests

nuomi2021:


nuomi2021:
@stone-d-chen, if we run it with `checkasm --test=vvc_deblock 1599970878`, you’ll notice that the vvc_h_loop_filter_chroma_10_strong_no_shift falls back to the weak filter. 
This happens primarily because the randomize_chroma_buffers did not account for dsam0, and dsam1, and also shift.

stone-d-chen:
> @stone-d-chen, if we run it with `checkasm --test=vvc_deblock 1599970878`, you’ll notice that the vvc_h_loop_filter_chroma_10_strong_no_shift falls back to the weak filter. This happens primarily because the randomize_chroma_buffers did not account for dsam0, and dsam1, and also shift.

Yeah this also occurred in the hevc checkasm where we copied it from originally, I haven't really looked into how to guarantee it, not sure if you feel it's necessary to have that guarantee?

I'll pull and look into the test failures on linux this weekend.

nuomi2021:
Yes, if possible. We should ensure strong filter tests and generate only strong filters. This will help us write assembly code and assist others in debugging issues.

PR #278, opened 2025-02-09
Deblock chroma: remove stack

nuomi2021:
With this refact, the chroma weak avx is 1.5x faster than c version. and it's no stack usage.
see https://github.com/ffvvc/FFmpeg/commit/31a0f2a419fb7b92021b078053bca592fe4f4d66

### DIFF
--- a/libavcodec/x86/vvc/vvc_deblock.asm
+++ b/libavcodec/x86/vvc/vvc_deblock.asm
@@ -1,4 +1,25 @@
-; from hevc_deblock.asm, grap all the tranpose macros
+; /*
+; * Provide SIMD deblock functions for VVC
+; * Copyright (c) 2025 Stone Chen
+; * Copyright (c) 2025 Nuo Mi
+; * All rights reserved.
+; *
+; * This file is part of FFmpeg.
+; *
+; * FFmpeg is free software; you can redistribute it and/or
+; * modify it under the terms of the GNU Lesser General Public
+; * License as published by the Free Software Foundation; either
+; * version 2.1 of the License, or (at your option) any later version.
+; *
+; * FFmpeg is distributed in the hope that it will be useful,
+; * but WITHOUT ANY WARRANTY; without even the implied warranty of
+; * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+; * Lesser General Public License for more details.
+; *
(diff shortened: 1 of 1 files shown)


libavcodec/x86/vvc/vvc_deblock.asm
```diff
@@ -66,18 +66,14 @@ ALIGN 16
...
     mova            m10, m1          ; save p2
     mova            m12, m2          ; save p1
 
     paddw           m14, m1, m2      ; p2 + p1
     paddw           m13, m3, m4      ; p0 + q0
     paddw           m13, m14         ; p2 + p1 + p0 + q0
 
-    cmp            no_pq, 0
-    je      .end_p_calcs
-    pand             m11, [rsp + 16] ; which p
```
stone-d-chen, line 79: Is it okay to remove no_p no_q?

stone-d-chen: Just kidding, I see the commit message now, I'll have to look up plt but LGTM.

stone-d-chen:
Is there a way to run the fuzz regression tests locally?
Otherwise this is much cleaner than my implementation, thank you for your refactor!

libavcodec/x86/vvc/vvc_deblock.asm
nuomi2021: If palette coding is used, the deblocking filter will be disabled.
With the current code, we apply weak -> one-side -> strong filtering. It should be easy to add no_q when needed.

nuomi2021:
> Is there a way to run the fuzz regression tests locally?

fuzz is not a problem. I merged new cases which not fixed by current deblock_asm branch. 
But window failure maybe a issue. Need to check why.

> Otherwise this is much cleaner than my implementation, thank you for your refactor!

Thank you for your great help with this! We still need to add the luma long filter code, and after that, I believe we’ll be ready to upstream. I’ll be dedicating more time to this in the coming weeks.

stone-d-chen:
> > Is there a way to run the fuzz regression tests locally?
> 
> fuzz is not a problem. I merged new cases which not fixed by current deblock_asm branch. But window failure maybe a issue. Need to check why.

Oh yeah, I just assumed it was the same problem. I've switched to windows and it seems like both yasm/nasm are failing checkasm and conformance.

PR #280, opened 2025-02-10
Fix windows

stone-d-chen:
same as https://github.com/nuomi2021/FFmpeg/pull/3 to check if runs pass

### DIFF
--- a/libavcodec/x86/vvc/vvc_deblock.asm
+++ b/libavcodec/x86/vvc/vvc_deblock.asm
@@ -350,14 +350,14 @@ cglobal vvc_h_loop_filter_chroma_8, 9, 11, 16, pix, stride, beta, tc, no_p, no_q
     sub           pix0q, src3strideq
     sub           pix0q, strideq
 
-    movq             m0, [pix0q             ]  ;  p3
-    movq             m1, [pix0q +   strideq]   ;  p2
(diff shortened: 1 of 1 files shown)


libavcodec/x86/vvc/vvc_deblock.asm
```diff
@@ -53,7 +53,7 @@ SECTION .text
     cmp         shiftd, 1
     jne             %%no_shift
 %if %3
-    movdqa          %1, [%2]
+    movu            %1, [%2]
```
nuomi2021: maybe movdqu is better. the dqu will not over read.

libavcodec/x86/vvc/vvc_deblock.asm
```diff
@@ -41,7 +41,7 @@ cextern pw_4096
 cextern pw_8192
 cextern pw_m1
 cextern pd_1
-cextern pd_3
+pd_3 :           times 4 dd  3
```
nuomi2021: after the refact pd_3 and some others are not needed. 
see https://github.com/ffvvc/FFmpeg/pull/278/files#diff-519e64dabcabdb9e30163dbbd4c91c2037fea9304dcbc2da1313fc6044c4f1c4R42

nuomi2021:
replaced by https://github.com/ffvvc/FFmpeg/pull/278
thank you for the pr

PR #281, opened 2025-02-12
vvc_deblock.c: fix RANDCLIP

stone-d-chen:
Previously RANDCLIP(x, diff) was computing `x - diff` and then clipping it between (0, max_pixel_val + rnd() % 2 * diff). This means we're not really generating a random value in the range.

Instead compute (x - diff) + rnd() % 2 * diff. This returns  a value such that abs(value - x) < diff.

This greatly improves the generation of strong deblocking data.

### DIFF
--- a/tests/checkasm/vvc_deblock.c
+++ b/tests/checkasm/vvc_deblock.c
@@ -67,8 +67,8 @@ static const uint32_t pixel_mask[3] = {0xffffffff, 0x03ff03ff, 0x0fff0fff};
     else \
         *(uint16_t*)(&x) = z; \
 } while (0)
-#define RANDCLIP(x, diff) av_clip(GET(x) - (diff), 0, \
-    (1 << (bit_depth)) - 1) + rnd() % FFMAX(2 * (diff), 1)
+#define RANDCLIP(x, diff) av_clip(GET(x) - (diff)  + rnd() % FFMAX(2 * (diff), 1), \
+0, (1 << (bit_depth)) - 1)
 
 static void randomize_params(int32_t *beta, int32_t *tc, const int is_luma, const int bit_depth, const int size)
 {


tests/checkasm/vvc_deblock.c
```diff
@@ -67,8 +67,8 @@ static const uint32_t pixel_mask[3] = {0xffffffff, 0x03ff03ff, 0x0fff0fff};
     else \
         *(uint16_t*)(&x) = z; \
 } while (0)
-#define RANDCLIP(x, diff) av_clip(GET(x) - (diff), 0, \
-    (1 << (bit_depth)) - 1) + rnd() % FFMAX(2 * (diff), 1)
+#define RANDCLIP(x, diff) av_clip(GET(x) - (diff)  + rnd() % FFMAX(2 * (diff), 1), \
```
nuomi2021, line 70: Hi Stone,
Do we need to update HEVC as well?
If so, could you submit the HEVC patch for upstream review first?

stone-d-chen: Yep sounds good!

stone-d-chen:
@nuomi2021 been looking into further improvements to the luma generation, it seems fairly non-trivial.

One of the main issues is occasionally `(d0 << 1) < beta_2` condition fails in the filter template. 

Where `d0 = abs(p2 - 2 * P1 + P0) +  abs(Q2 - 2 * Q1 + Q0)`

The current code does actually try to compensate for it, since (d0 << 1) < beta_2  == d0 < (beta_2 >> 1) which is beta_3.

It becomes difficult to solve both constraints while also satisfying (d0 + d1 < beta).

I attempted to put it into a computer algebra solver (wxMaxima) but it's quite messy.

nuomi2021:
If it's difficult, perhaps we should approach it as it is.
I'll rebase the code to the latest version to fix the fuzz issue.
Then, we can work together on two tasks:

1. Enabling a larger filter for luma—this is the last missing part.
2. Enabling AVX2—this will further improve performance.

Which one do you prefer?

stone-d-chen:
> If it's difficult, perhaps we should approach it as it is. I'll rebase the code to the latest version to fix the fuzz issue. Then, we can work together on two tasks:
> 
> 1. Enabling a larger filter for luma—this is the last missing part.
> 2. Enabling AVX2—this will further improve performance.
> 
> Which one do you prefer?

AVX2 sounds good, we need to modify the C side to expose multiple blocks right? I'm trying to learn more about video decoding overall some more exposure to the c would be good.

nuomi2021:
> > If it's difficult, perhaps we should approach it as it is. I'll rebase the code to the latest version to fix the fuzz issue. Then, we can work together on two tasks:
> > 
> > 1. Enabling a larger filter for luma—this is the last missing part.
> > 2. Enabling AVX2—this will further improve performance.
> > 
> > Which one do you prefer?
> 
> AVX2 sounds good, 

👍

> we need to modify the C side to expose multiple blocks right? 

Yes, we need to set up parameters for a single line within a CTU.  SSE can process 16 bytes at a time, AVX2 can handle 32 bytes, and AVX-512 can manage 64 bytes per operation.

> I'm trying to learn more about video decoding overall some more exposure to the c would be good.

You can start from https://www.amazon.com/Coding-Video-Practical-Guide-Beyond/dp/1118711785 :)

